PR #157, opened 2018-01-17
Update group sample response to include group status

nicolasdelfino:
We noticed that group status is returned in list all groups endpoint.
This was not reflected in the sample response.


retrovampire:
Thanks so much for this! Great to see the community growing :)

PR #518, opened 2019-11-14
API-4753 Add API changelogs

kogierman:
Add changelogs for Configuration API v3, Agent Chat API and Customer Chat API


ale-kacperczyk:
No longer needed. Changelogs were moved to feature/dps-1165. The content for v3.2 will be added there too.

PR #559, opened 2020-01-23
API-5443: Add info about 'id' in group_chooser options

Adimus11:


ale-kacperczyk:
created a clone branch: _feature/PI-5443-id-in-group-chooser_ to have a feature branch.

PR #562, opened 2020-01-24
API- 5443 ID in group_chooser

ale-kacperczyk:
Copy of the branch: _API-5443-update-docs-about-id-in-group-chooser_ made to have a feature branch.

There's a new parameter `id` in the **Form field** data structure (`fields.answer.id` for the `group_chooser` field type).  It's a unique identifier for each option Customers can choose. This `id` is also returned by the **Get Form** method.

### DIFF
--- a/content/messaging/agent-chat-api/v3.2/index.mdx
+++ b/content/messaging/agent-chat-api/v3.2/index.mdx
@@ -315,6 +315,7 @@ The **Filled form** event contains data from a form (prechat or postchat survey)
 		"id": "154417206262605324",
 		"label": "Choose department",
 		"answer": {
+      "id": "2",
 			"group_id": 1,
 			"label": "Marketing"
 		}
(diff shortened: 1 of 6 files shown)


content/messaging/agent-chat-api/v3.2/index.mdx
```diff
@@ -1031,7 +1032,7 @@ A component of the [Filled form](#filled-form) event.
 | `id`                    | yes      | `string`                | Field id, for all field types                                                                            |
 | `label`                 | yes      | `string`                | Field label; for all field types                                                                         |
 | `answer`                | yes      | `any`                   | For all field types                                                                                      |
-| `answer.id`             | yes      | `string`                | Answer id; for all field types                                                                           |
+| `answer.id`             | yes      | `string`                | A unique identifier of each option a Customer can choose. For all field types.                           |
```
Adimus11: Everything is perfect, but  I just wondering if `unique` is good here, because `id` is unique only among the answers list for the given question.

ale-kacperczyk: @Adimus11 So you're saying that this `id` won't be unique when out of context of `answer`? That seems reasonable. So maybe we should simply go for `An identifier...`

Adimus11: @ale-kacperczyk Sounds reasonable 👍

PR #566, opened 2020-01-27
(NO-DPS) Add filter to hide articles

jakubsikora:


content/index.mdx
```diff
@@ -3,6 +3,7 @@ weight: 0
 title: "LiveChat Platform Docs"
 tagline: "The source of knowledge of the LiveChat Platform"
 desc: "LiveChat is an online customer service software with live support, help desk, and web analytics capabilities. Browse the LiveChat Platform documentation to find out more."
+hidden: false
```
jakubsikora: This is needed so graphql has some default value. It will break the build without it.

PR #573, opened 2020-02-07
[DPS-1583 ] Event tracking

ale-kacperczyk:
Events for:

- clicking a category from the top menu
- clicking the  link to Developer Console
- search results (what was typed: input, which suggestion was chosen: suggestion.url)
- selected docs version from the dropdown

+ fixed one tiny thing in `PopperTooltip`


src/components/core/header.js
```diff
@@ -90,20 +91,23 @@ const activeLinkStyle = color => ({
...
-        {label}
-      </Link>
-    )}
+        <Link
+          to={`/${slug}/`}
+          partiallyActive
+          css={linkStyle}
+          activeStyle={activeLinkStyle(color)}
+          onClick={() => logAmplitudeEvent('Tab selected from top menu', {
+            slug: slug
```
jakubsikora:
```suggestion
            slug
```

You can use object shorthand notation

src/utils/index.js
```diff
@@ -1,3 +1,19 @@
+import amplitude from "amplitude-js";
+
+export const setupAmplitude = () => {
+  if (typeof window === "undefined") {
+    return;
+  }
+
+  amplitude.getInstance().init("656f6573a88e63d3dc2ce32e33ef8284");
```
fjaskolski: I'd suggest moving this id to an environment variable in `.env` file. This way we can set up separate Amplitude instances for prod and labs environments.

Please see the following article: https://www.gatsbyjs.org/docs/environment-variables/

jakubsikora: This will now use production amplitude key for all envs. We should start using .envs files. You could have a look at https://www.gatsbyjs.org/docs/environment-variables/, or if you need help I'm happy to help.

src/components/core/header.js
```diff
@@ -90,20 +91,23 @@ const activeLinkStyle = color => ({
 const MenuElement = ({ label, href, slug, color, ...props }) => (
   <MenuElementWrapper {...props}>
     {href ? (
-      <a href={href} css={linkStyle} target="_blank" rel="noopener noreferrer">
+      <a href={href} css={linkStyle} onClick={() => logAmplitudeEvent('Link to Developer Console clicked', {})} target="_blank" rel="noopener noreferrer">
```
fjaskolski: The `<MenuElementWrapper/>` component aimed to be generic. I'd suggest rephrasing the event name from  `Link to Developer Console clicked` to `External link clicked` with `href` passed as an additional event param.

src/utils/index.js
```diff
@@ -9,7 +25,20 @@ export const setupDocsearch = () => {
...
     indexName: "livechatinc",
     inputSelector: "#search",
-    debug: false
+    debug: false,
+    handleSelected: function (input, event, suggestion, datasetNumber, context) {
+      if (
+        context.selectionMethod === "click" ||
+        context.selectionMethod === "enterKey"
+      ) {
+        logAmplitudeEvent("User selected suggestion and input", {
```
fjaskolski: I'd suggest using a standardized pattern for naming events. For instance: 
- always use present tense (it's shorter and a bit easier to read) or always use past tense (longer, but might be less confusing),
- don't include the subject (it's already defined implicitly by the way it's displayed in the analytics panels),
- stick to the same convention throughout the entire project (it's easier to guess the name of the event).

There are multiple approaches to pick from: 

- https://help.amplitude.com/hc/en-us/articles/115000465251#how-should-i-name-my-events
- https://segment.com/academy/collecting-data/naming-conventions-for-clean-data/
- https://supermetrics.com/blog/naming-conventions-google-analytics

src/utils/index.js
```diff
@@ -9,7 +25,20 @@ export const setupDocsearch = () => {
...
+          url: suggestion.url,
+          input: input.getVal()
+        });
+
+        window.location.href = suggestion.url;
```
fjaskolski: This would cause the page to reload, and possibly the event not being sent (not sure if Amplitude SDK can handle such cases and safely finish sending the event after the reload).

I'd be curious if we could use [Gatsby navigate](https://www.gatsbyjs.org/docs/gatsby-link/#how-to-use-the-navigate-helper-functionc) instead. It would also provide better experience for the user.

src/utils/index.js
```diff
@@ -9,7 +25,20 @@ export const setupDocsearch = () => {
     apiKey: "f53a424573adab20d04faa2db150c349",
```
fjaskolski: Once we have a `.env` file, it would be great to move this public API key over there as well.

src/utils/index.js
jakubsikora: As far as I know the page always reload on search there.

src/utils/index.js
fjaskolski: Totally forgot this comes from [Algolia DocSearch](https://docsearch.algolia.com/) and we don't have a labs equivalent. Please feel free to paste the same key for both prod and labs environments.

src/utils/index.js
ale-kacperczyk: The events are being sent : )

ale-kacperczyk: Changed to Gatsby navigate

walaszczykm:
Hello 👋just to note, you have merged 3 files with no EOLs:
- `.env.development`
- `.env.labs`
- `.env.production`

PR #581, opened 2020-02-19
New param image_original_url in Get URL details

ale-kacperczyk:
Method **Get URL Details** (Customer API, Web & RTM) returns a new parameter, `image_original_url`.

### DIFF
--- a/content/messaging/customer-chat-api/v3.2/index.mdx
+++ b/content/messaging/customer-chat-api/v3.2/index.mdx
@@ -2615,6 +2615,13 @@ It returns the info on a given URL.
 | --------- | -------- | --------- | ----------------- |
 | `url`     | Yes      | `string`  | Valid website URL |
 
+#### Response
+
+| Parameter           | Data type | Notes             |
+| ------------------- | --------- | ----------------- |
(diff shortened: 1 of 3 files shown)


content/messaging/customer-chat-api/v3.2/index.mdx
```diff
@@ -2615,6 +2615,13 @@ It returns the info on a given URL.
 | --------- | -------- | --------- | ----------------- |
 | `url`     | Yes      | `string`  | Valid website URL |
 
+#### Response
+
+| Parameter           | Data type | Notes             |
+| ------------------- | --------- | ----------------- |
+| `image_url`         | `string`  | URL of the minified image hosted on the LiveChat's CDN |
```
lwojciechowski: There are also other parameteres: `title`, `description`, ... Shouldn't we describe them all?

ale-kacperczyk: Yes and no. Usually, when response objects are self-explanatory we don't describe them unless there's some extra info. Is there any extra info regarding all those undescribed response objects, something that isn't obvious but might be useful?

PR #592, opened 2020-03-04
Feature/pp 2122 operators

skix123:


content/monetization/partner-program-api/index.mdx
```diff
@@ -1714,6 +1720,87 @@ curl --request GET \
...
+- `operators` - **required** - list of the **users** you want to invite to the license.
+
+Struct of **user** object:
+
+- `email` - **required** - user email
+- `role` - **required** - `administrator` or `agent` ([learn more about the roles](https://www.livechatinc.com/help/livechat-roles-owner-administrators-and-agents/))
+
+#### Response
+
+- `200 - OK` - success (could be partly, check for `status_code` in response body), see the example on the right
```
ale-kacperczyk: What does it mean that success could be partial?

ale-kacperczyk:
```suggestion
- `200 - OK` - success (could be partial, check for `status_code` in the response body), see the example on the right
```

skix123: It means that some users might be invited and some not - e.g. because of already taken email

PR #601, opened 2020-03-13
API-5352 agent login away

gluszczyk:


content/messaging/agent-chat-api/changelog/index.mdx
```diff
@@ -10,30 +10,37 @@ desc: "Changelog of the Agent Chat API"
...
 - The method **Get Chat Threads** ([Web](/messaging/agent-chat-api/v3.2/#get-chat-threads) & [RTM](/messaging/agent-chat-api/v3.2/rtm-reference/#get-chat-threads)) now requires `thread_ids`. Also, it no longer returns `threads_summary` in the response.
 - The [**Property**](/messaging/agent-chat-api/v3.2/#properties) data structure now has the same format in requests as in responses and pushes.
 - There's a new parameter `id` in the [Form field](/messaging/agent-chat-api/v3.2/#form-fields) data structure (`fields.answer.id` for the `group_chooser` field type).  It's an identifier for each option Customers can choose.
 - The **Add User to Chat** method ([Web](/messaging/agent-chat-api/v3.2/#add-user-to-chat) & [RTM](/messaging/agent-chat-api/v3.2/rtm-reference/#add-user-to-chat)) accepts an optional parameter `require_active_thread`.
+- There's a new parameter `away` in the [Login](/messaging/agent-chat-api/v3.2/rtm-reference/#login) request. Controls whether the connection should be considered to be in the away state by the auto-away feature. Setting it to true should help avoiding those pesky chat assignments that happen right after random reconnections.
```
kogierman: I'm not so sure about the wording here, "should help" doesn't sound very confident :D but let's here what Ola has to say

ale-kacperczyk: My thoughts:

1. I wonder if Changelog is the right place to describe the role of this parameter in detail. WDYT?  What's the pattern in other docs? My intuition tells me we should only highlight its role and provide more details in the desc of the Login method.
2. Yeah, _should help_ sounds like we're not sure if it works. `(...) helps avoiding...` 
3. I would replace pesky with something more common, like troublesome.

All in all, I suggest adding this to Changelog:

- There's a new parameter `away` in the [Login](/messaging/agent-chat-api/v3.2/rtm-reference/#login) request. It was introduced to avoid problematic chat assignments that happen right after random reconnections.

We could explain the rest in the method desc.

gluszczyk: I'm ok with the new wording.

content/messaging/agent-chat-api/v3.2/rtm-reference/index.mdx
```diff
@@ -3693,25 +3693,26 @@ It returns the initial state of the current Agent.
 | `application`                       | No       | `object`  |                                                                                                                                  |
 | `application.name`                  | No       | `string`  | Application name                                                                                                                 |
 | `application.version`               | No       | `string`  | Application version                                                                                                              |
+| `away`                              | No       | `bool`    | Controls whether the connection should be considered to be in the away state by the auto-away feature. Defaults to false.        |
```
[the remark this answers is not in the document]
ale-kacperczyk: We could describe the auto-away feature in the description of **Set Away Status** and link to it. For example:

## Set Away Status

Sets the requesting Agent's status to `not_accepting_chats`.

**Auto-away**
With the auto-away feature, Agents can set up an inactivity period after which their statuses automatically switch to **Don't accept chats** (`not_accepting_chats`). This way, Agents won't be assigned new chats while they're not at their desks. 

There's a problematic edge case you may have come across. If an Agent was auto-assigned the `not_accepting_chats` status, but in the meantime, they were disconnected and logged in again, they can still be assigned some chats. To solve this issue, Agents can [**Login**](#login) with `away: true`. In other words, they log in with the status `not_accepting_chats` already set. It prevents assigning chats to them even after they reconnect.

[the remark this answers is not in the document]
ale-kacperczyk: tell me what you think 🙏🏻 too long?

PR #608, opened 2020-03-20
Describe API versioning

ale-kacperczyk:
Described API versioning in **Messaging Overview > APIs Overview > API versioning** and in the **Versioning** section of each API document.

**feature branch:** https://developers.labs.livechat.com/docs/feature/versioning

### DIFF
--- a/content/management/configuration-api/index.mdx
+++ b/content/management/configuration-api/index.mdx
@@ -14,7 +14,7 @@ Configuration API is a service for storing configuration of license. You can set
 
 ## Versioning
 
-This document describes the **Configuration API v3.1**. [Read more about versionig...](/messaging/#api-versioning)
+This document describes the **Configuration API v3.1**. This is the latest stable version recommended for the production use. [Read more about versionig...](/messaging/#api-versioning)
 
 ## Authentication
 
@@ -44,7 +44,6 @@ Bot Agents enables writing integrations using the **Agent Chat API** - both [RTM
 A Bot Agent shares the SSO access token with the Agent who created the Bot. Each Bot Agent is a resource owned by an application in Developers Platform, identified by its own `client_id`. 
 
 Unlike Agents, Bot Agents don't have passwords or emails - you cannot log in as a Bot. 
-
 
 ## Methods
 
--- a/content/management/configuration-api/index.mdx
+++ b/content/management/configuration-api/index.mdx
@@ -14,12 +14,11 @@ Configuration API is a service for storing configuration of license. You can set
 
 ## Versioning
 
-This document describes the **Configuration API v3.1**.
+This document describes the **Configuration API v3.1**. [Read more about versionig...](/messaging/#api-versioning)
 
 ## Authentication
 
 Authentication for Configuration API is handled by access tokens. Find out how to get an access token from [Agent authorization flows](/getting-started/authorization/#agent-authorization-flows). If a method requires particular authorization scopes, you’ll find them included in the method description. Each request should contain the `Authorization: Bearer <your_access_token>` header.
-
 
 ## Data centers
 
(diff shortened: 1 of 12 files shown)


content/messaging/index.mdx
```diff
@@ -88,6 +88,30 @@ To learn about differences between these two API types, see the comparison below
 
 If you're not sure, which implementation to choose, we suggest reading about particular APIs. You'll find the **When to use RTM/Web API** section in each document - refer to the table above for links.
 
+## API Versioning
+
+LiveChat Messaging API doesn't follow <a href="hhttps://semver.org/" target="_blank">semantic versioning</a>. We suggest you learn our system and terminology:
+
+- **stable**
+  
+  The latest stable version we encourage developers to use. We don't introduce any new functionalities or breaking changes - only bug fixes are allowed.
```
bkubiak: "are allowed" sounds a little bit funny here ;) bug fixes are expected ;)

content/messaging/index.mdx
```diff
@@ -88,6 +88,30 @@ To learn about differences between these two API types, see the comparison below
 
 If you're not sure, which implementation to choose, we suggest reading about particular APIs. You'll find the **When to use RTM/Web API** section in each document - refer to the table above for links.
 
+## API Versioning
+
+LiveChat Messaging API doesn't follow <a href="hhttps://semver.org/" target="_blank">semantic versioning</a>. We suggest you learn our system and terminology:
```
bkubiak: I'm not sure if we should write about not following semver :/ It can sound negative, I don't know.

content/messaging/index.mdx
```diff
@@ -88,6 +88,30 @@ To learn about differences between these two API types, see the comparison below
...
+
+- **stable**
+  
+  The latest stable version we encourage developers to use. We don't introduce any new functionalities or breaking changes - only bug fixes are allowed.
+
+- **developer preview (dev preview)**
```
bkubiak: perhaps just one name is enough, wdyt?

content/management/configuration-api/index.mdx
```diff
@@ -14,13 +14,12 @@ Configuration API is a service for storing configuration of license. You can set
 
 ## Versioning
 
-This document describes the **Configuration API v3.1**.
+This document describes the **Configuration API v3.1**. [Read more about versionig...](/messaging/#api-versioning)
```
bkubiak: Why do we have different description here? How about adding:
>This is the latest stable version recommended for the production use. [Read more about versionig...](/messaging/#api-versioning)

content/messaging/index.mdx
ale-kacperczyk: Is it a bad thing we're not following it? If it's not, I would leave this statement. At least we're transparent and we don't mislead developers in this area.

content/messaging/index.mdx
ale-kacperczyk: Haven't thought about it but you're right :)

content/messaging/index.mdx
ale-kacperczyk: I'd like to have this abbreviation explain somewhere in the docs. It's quite obvious that developer preview = dev preview, but since we used the two options interchangeably, let's keep it here for the sake of clarity.

content/messaging/index.mdx
bkubiak: ok

bkubiak: I mean few apis follow semver, it was introduced mainly for libraries or applications, not online apis

content/messaging/index.mdx
bkubiak: ok

content/messaging/index.mdx
ale-kacperczyk: @Andarist, what's your opinion on this?

Andarist: @bkubiak is right that platform APIs usually do not adhere to semver. Ultimately it doesn't matter much, the important part is to describe the versioning policy accurately. We don't have to mention semver explicitly, but definitely we need to mention that breaking changes might get introduced between version (which we do). Maybe a good thing that would help to emphasize this would be a short paragraph about API upgrades? Like - when you should do it, what you should check when doing so etc.

A stable~ release schedule would be also helpful for developers, something like in case of Shopify and others:

ale-kacperczyk: thank you! I suggest we kick off the semver mention, publish the content as it is, and develop on upgrades in the second iteration. Some visuals would be great, I can arrange for that to be taken care of.

PR #617, opened 2020-03-30
Rename agent/customer names in examples.

gluszczyk:


gluszczyk:
> Awesome 😎We could also adjust emails in the future

sure thing, but that might be more complicated as we need to avoid real domains
...unless we create accounts for them in the LiveChat domain, as aliases for docs/developers platform email?

ale-kacperczyk:
Wouldn't thomas-anderson@example.com do the job? Great idea with emails on our domain, especially for the Agent

gluszczyk:
ok, we can start with `example.com`
for agent let's go with smith@example.com, as `agent` is not a name
and for customer `thomas-anderson` or `t.anderson`?

ale-kacperczyk:
`t.anderson` as it's shorter :)

content/messaging/agent-chat-api/v3.2/rtm-reference/index.mdx
```diff
@@ -5315,7 +5315,7 @@ Informs about the message a Customer is currently typing. The push payload conta
   "chat_id": "PJ0MRSHTDG",
   "thread_id": "K600PKZON8",
   "sneak_peek": {   
-    author_id: "d17cd570-11a9-45c0-45c0-1b020b7586dc",
+    author_id: "b7eff798-f8df-4364-8059-649c35c9ed0c",
```
ale-kacperczyk, line 5318: Why isn't it a full email here?

gluszczyk: sneak peeks are generated by customers so `author_id == customer_id`

PR #618, opened 2020-03-30
remove vistiorSDK from the menu and inform that it is deprecated

klarzynskik:
remove vistiorSDK from the menu and inform that it is deprecated


klarzynskik:
@ale-kacperczyk apologies, was not aware of such marvelous feature, thanks for pointing it out

fixed

PR #621, opened 2020-03-31
use correct JSX attributes in the StarIcon

klarzynskik:
while updating the VisitorSDK I've noticed warnings regarding invalid DOM properties for the `StarIcon`


ale-kacperczyk:
I think we those warnings, don't we @jakubsikora ? 😂

PR #633, opened 2020-04-14
Add info about chats without any threads in 'list_chats'

Adimus11:


ale-kacperczyk:
Please, create a copy of this branch that starts with `feature/` so that I can check it out on labs

PR #649, opened 2020-05-07
DPS direct charge update + allow developer to set recurring charge in months time frame

sokool:
## 🚀 Links

- [DPS-1895](https://livechatinc.atlassian.net/browse/DPS-1895)
- [DPS-1839](https://livechatinc.atlassian.net/browse/DPS-1839)

## 📓 Description
- allow developer to set recurring charge in months time frame
- direct charge update implementation


ale-kacperczyk:
Moved to #650

PR #684, opened 2020-06-16
Add info that greeting's addon property is optional

Andarist:
This is a newly introduced property, just in the preview version so not much to change besides adding this information.


ale-kacperczyk:
When `addon` was being added, I asked not to add those "JSON" comments. The reason is that I'm trying to move all JSONs (pushes, responses, etc) into separate JSON files, and then import them and pass to components. So, I will have to delete them anyway.

We could add info about it being optional in the table below, WDYT?

PS Please, create a feature branch next time: `feature/` 🙏🏻

PR #700, opened 2020-06-23
Describe new moments methods: setExternalLink and sendSystemMessage

klarzynskik:
## 🚀 Links

- [Feature branch](https://developers.labs.livechat.com/docs/feature/new-moments-capabilities/)
- [Jira](https://livechatinc.atlassian.net/browse/CW-1580)

## 📓 Description
Describe new moments methods: `setExternalLink` and `sendSystemMessage`

### DIFF
--- a/content/extending-ui/extending-chat-widget/chat-widget-moments/index.mdx
+++ b/content/extending-ui/extending-chat-widget/chat-widget-moments/index.mdx
@@ -103,7 +103,7 @@ and then import `createMomentsSDK` function which is exported as default from pa
 
 ```js
 import createMomentsSDK from "@livechat/moments-sdk";
-createMomentsSDK({ title: "My App" }).then(momentsSDK => {
+createMomentsSDK({ title: "My App" }).then((momentsSDK) => {
   // your code
 });
 ```
@@ -131,10 +131,25 @@ window.onMomentAsyncInit = function(createMomentsSDK) {
 Sends a message as a visitor in currently active chat.
 Parameters:
 
-- text - Message text
+- `text` - Message text
 
 ```js
 momentsSDK.sendMessage({ text: "Chosen date: 19.12.2009" });
+```
+
+### sendSystemMessage
+
+Sends a system message in the currently active chat.
+Parameteres:
+
+- `text` - Message text
+- `recipients` - Optional, defines those who can display the message: `all` (default) or `agents`
+
+```js
+momentsSDK.sendSystemMessage({
+  text: "Survey submitted",
+  recipients: "agents",
+});
 ```
 
 ### setAttributes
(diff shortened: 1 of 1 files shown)


content/extending-ui/extending-chat-widget/chat-widget-moments/index.mdx
```diff
@@ -182,17 +208,14 @@ Alternatively you can setup the [MessageBox](/extending-ui/extending-agent-app/a
...
+## Changelog
 
-- Add Calendly app example
+### [v1.1.0] - 2019-12-17
 
-#### [v0.0.2] - 07.08.2019
+#### Added
 
-- Update methods description with UMD build usage.
-- Describe MessageBox integration as a place to use Moments.
```
walaszczykm, line 194: Why you have removed those version from the changelog?

klarzynskik: because I think that the `moments-sdk` should be versioned, not the documentation itself, therefore I've updated that and changed the nesting to include that

walaszczykm: The documentation itself should version and we are doing this in other docs. You should know when and what has changed in the documentation itself.

klarzynskik: I agree that we should keep versioning here, but I don't think that's valuable information that we've added the `calendly app example`, or `described MessageBox integration`

This is not done this way in any of other documentations, and is unrelated to the actual version of the `moments-sdk`

It's valuable to keep important things like change of the shape, deprecated or added methods, and so on, but the examples being added or removed is unrelated to the actual package version, and because of that the changelog is now nested in the `Moments SDK`, not the `Moments` themselves

WDYT?

walaszczykm: ok discussed f2f we should keep the changelog in syn with the changes in the package not the content of the doc itself 👍

content/extending-ui/extending-chat-widget/chat-widget-moments/index.mdx
```diff
@@ -131,23 +131,49 @@ window.onMomentAsyncInit = function(createMomentsSDK) {
...
+- `customProperties` - Customer's additional data object
 
 ```js
 momentsSDK.setAttributes({ clientId: "SJ3CJ6JVNMK42A", source: "Organic" });
 ```
 
+### setExternalLink
+
+Sets the external link for the fullscreen version of the app, which will be opened in the new tab.
+Arguments:
```
konradk, line 169: In `sendSystemMessage` description we use `Parameteres`

PR #705, opened 2020-06-26
Flow with integrations

ale-kacperczyk:
## 🚀 Links

- [Feature branch](https://developers.labs.livechat.com/docs/feature/dps-2063)
- [Jira](https://livechatinc.atlassian.net/browse/DPS-2063)

## 📓 Description
- Add a graphic and description for the flow with integration.
- Moved the sections around
- Added a redirect for (#the-flow) which changed to (#flow-between-services)

### Content

- New content
  
- Functional change (a new redirect)


content/messaging/index.mdx
```diff
@@ -65,15 +77,13 @@ The first connection allows for sending an event, for example, a message, by an
 
 Apart from pushes, actions can also trigger the LiveChat's backend to generate webhooks. To receive a webhook, you need to [register it](/management/configuration-api/#register-webhook) first. That's how you let the LiveChat's backend know you want to be notified about certain actions. The LiveChat's backend will send webhooks to the target webhook URL you specify during webhook registration.
 
-## Chat routing
-
-Routing is the process of **assigning chats to agents**. The primary goal of the routing mechanism is to distribute chats to all available agents who use the same license.
+### Flow with integrations
```
bkubiak, line 80: This flow doesn't apply to all integrations but only to those that enable new communication channel

ale-kacperczyk: That's true. Do you think this headline is misleading and with should change it or is it okay if we highlight that in the description? For example

> ### Flow with integrations
> Developers can build integrations that extend LiveChat's messaging capabilities, for example, via a new communication channel. Consider the example of the integration app that connects LiveChat with Facebook Messenger. Keep in mind the flow presented in the diagram below is typical for a new communication channel and it would be different for an app that simply extends the LiveChat UI.

ale-kacperczyk: Why do you suggest to remove webhooks from the basic flow? Because it's not basic as you have to register webhooks first? Well, it's additional info, but we're giving a more full overview. I'd say we leave it for now, create a diagram dedicated to webhooks, put in the Webhooks section, and then simplify the basic flow. WDYT?

bkubiak: It's ok

content/messaging/index.mdx
```diff
@@ -65,15 +77,13 @@ The first connection allows for sending an event, for example, a message, by an
...
 
-Chats are being assigned to agents **either automatically or manually**. Our default routing mechanisms are described in [Understanding chat routing](https://www.livechatinc.com/help/understanding-chat-routing/).
+Developers can build integrations that extend LiveChat's messaging capabilities, for example, via a new communication channel. Consider the example of the <a href="https://www.livechat.com/marketplace/apps/facebook-messenger/" target="_blank">integration app</a> that connects LiveChat with Facebook Messenger.
```
ale-kacperczyk, line 82:
```suggestion
Developers can build integrations that extend LiveChat's messaging capabilities, for example, via a new communication channel. Consider the example of the <a href="https://www.livechat.com/marketplace/apps/facebook-messenger/" target="_blank">integration app</a> that connects LiveChat with Facebook Messenger. Keep in mind the flow presented in the diagram below is typical for a new communication channel and it would be different for an app that simply extends the LiveChat UI.
```

content/messaging/index.mdx
bkubiak: webhooks belong to integration flow more than to basic flow, but we can do it later

PR #708, opened 2020-06-30
Align request with response in get_group

janowsiany:
## 🚀 Links

- [Feature branch](https://developers.labs.livechat.com/docs/feature/get_group-align-request-response)


ale-kacperczyk:
What will it return if you only request with _id_?

janowsiany:
`{ "id": 19, "name": "Sports shoes", "language_code": "en" }`

PR #709, opened 2020-06-30
Update Chat Widget snippet code to the newest version

klarzynskik:
## 🚀 Links

- [Feature branch](https://developers.labs.livechat.com/docs/feature/update-widget-snippet-code/)

## 📓 Description
Update the Chat Widget snippet code to the newest version.

### DIFF
--- a/src/components/core/seo.js
+++ b/src/components/core/seo.js
@@ -62,13 +62,9 @@ function SEO({ desc, lang, meta, keywords, title }) {
           >
             <script type="text/javascript">
               {`
-    window.__lc = window.__lc || {};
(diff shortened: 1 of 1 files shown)


src/components/core/seo.js
```diff
@@ -62,13 +62,9 @@ function SEO({ desc, lang, meta, keywords, title }) {
           >
             <script type="text/javascript">
               {`
```
konradk, line 64: is {` required?

klarzynskik: yes, sir, I believe it is, `react-helmet` requires that to be wrapped

PR #727, opened 2020-07-20
Method description for List Bots

ale-kacperczyk:
## 🚀 Links

- [Feature branch](https://developers.labs.livechat.com/docs/feature/edit-list-bots)

## 📓 Description
Added a detailed method description explaining how the response may differ in different situations.

## ✅ Checklist (for API docs)

- v3.3 Configuration API ✅
- v3.2 Configuration API 

### Content

- Proofreading/content fixes

### DIFF
--- a/content/management/configuration-api/v3.3/index.mdx
+++ b/content/management/configuration-api/v3.3/index.mdx
@@ -888,10 +888,7 @@ curl -X POST \
 
 ### List Bots
 
-Lists all the Bots within a license that were created by the requester. The method behavior differs depending on who created the Bot and who the requester is.
-
-- If a Bot was created **directly by an Agent** (PAT authorization), all Agents within the license will get that Bot in the response, even with the `all` parameter set to `false`.
-- In a situation when a Bot was created by an Agent **via an application**, all Agents within a license will get the Bot in the response if they request it via that application. If the Agents, including the Bot creator, want to access it via a different application, they need to specify `all:true`.
+Returns the list of Bots created within a license. The method behavior differs depending on who created the Bot and what authorization method is used for the call.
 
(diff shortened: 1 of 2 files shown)


content/management/configuration-api/v3.3/index.mdx
```diff
@@ -888,18 +888,23 @@ curl -X POST \
 
 ### List Bots
 
+Lists all the Bots within a license that were created by the requester. The method behavior differs depending on who created the Bot and who the requester is.
```
litwinow, line 891: I think `...that were created by the requester` is misleading but I don' know how to rephrase it.
This method returns all bots created via given application if `all` is set to `false` and all bots on the license otherwise, but it's not possible to get all bots created by the requester.

content/management/configuration-api/v3.3/index.mdx
```diff
@@ -888,18 +888,23 @@ curl -X POST \
...
+
+- If a Bot was created **directly by an Agent** (PAT authorization), all Agents within the license will get that Bot in the response, even with the `all` parameter set to `false`.
```
litwinow, line 893: all agents will get that bot if they use PAT; if they use token from any app, they won't see the bot

content/management/configuration-api/v3.3/index.mdx
ale-kacperczyk: Hm, my idea was to represent either an agent or an application as a requester, but it's not very intuitive. Maybe, we should make the first sentence more general:

Returns the list of Bots created within a license.

litwinow: 👏 sounds good

PR #758, opened 2020-09-11
require more scopes in login due to information in pushes (all versions)

bkubiak:
## 🚀 Links

- [Feature branch](https://developers.labs.livechat.com/docs/feature/API-7147-login-scopes)
- [Jira](https://livechatinc.atlassian.net/browse/API-7147)

## 📓 Description

We need to require more scopes in `login` due to information in pushes.

In near future (and v3.3 only), we will provide a way to choose subscription on specific pushes so required scopes will depend on the request payload.


bkubiak:
@ale-kacperczyk can we deploy it to production?

PR #768, opened 2020-10-13
Fix hide example to use proper callback in old JS API docs

klarzynskik:
## 📓 Description
Fix the old JS API example, so it uses the proper callback


ale-kacperczyk (approved):
You can merge whenever you're ready :)

PR #831, opened 2021-02-03
Remove information about filters object which is no longer required i…

kacperf531:
…n v3.3 when providing a page ID.

## 📓 Description
In v3.3, filters requested on first page are remembered when paginating. Providing `filters` object in the request with `page_id` would result in validation error. note: I'm not sure if this should be included in the changelog, as this has been in place for a while, probably someone just overseen this detail.

## ✅ Checklist (for API docs)

- Changelog
- API versions
- Web & RTM
- **Table:** Available methods/webhooks/pushes
- **Table:** Specifics

For more guidelines, see [Updating API docs](https://livechatinc.atlassian.net/wiki/spaces/PAT/pages/761529400/Updating+API+docs).

### Content

- Proofreading/content fixes


ale-kacperczyk:
Please, change this branch's name to `feature/Remove-filters-info-in-3-3-for-pagination` so that's there's a labs preview.

PR #840, opened 2021-02-15
Describe customers:own, add and fix links

ale-kacperczyk:
## 🚀 Links

- [Feature branch](https://developers.labs.livechat.com/docs/feature/fixes-in-authorization)

## 📓 Description
Reported on Slack

### Content

- Proofreading/content fixes

### DIFF
--- a/src/constant/index.js
+++ b/src/constant/index.js
@@ -207,7 +207,7 @@ const SCOPES = [
   },
   {
     id: "customers:own",
-    description: "Permission for owning and managing customer identities. It allows for creating new customers via the Customer Chat API.",
+    description: "Permission for owning and managing customer identities. It allows for acquiring a customer token (for both existing and new customers) and using it to call the Customer Chat API as a customer.",
     role: "administrator",
     role_type: "secondary",
     product: "LiveChat",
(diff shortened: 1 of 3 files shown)


src/constant/index.js
```diff
@@ -207,7 +207,7 @@ const SCOPES = [
   },
   {
     id: "customers:own",
-    description: "Permission for managing customers identities",
+    description: "Permission for owning and managing customer identities. It allows for creating new customers via the Customer Chat API.",
```
ale-kacperczyk, line 210: @bkubiak

src/constant/index.js
```diff
@@ -221,7 +221,7 @@ const SCOPES = [
   },
   {
     id: "customers:rw",
-    description: "Read/write permission for customers",
+    description: "Read/write permission for existing customers. It allows for creating new customers via the Agent Chat API.",
```
ale-kacperczyk, line 224: @bkubiak

bkubiak: "It allows *also*..."?

But I don't insist :)

src/constant/index.js
bkubiak: I have some doubts regarding "via the Customer Chat API." - in customer-api there's no method for creating a customer explicitly. You just use their identity (impersonate them) and login, start_chat, etc (this would automatically create a new customer if they didn't exist).

The part regarding customer creation an equivalent of customers:rw in agent-api.

PR #868, opened 2021-03-22
API-8565: rename reorder_ to update_auto_access

lpmichalik:
drop `customer_` from rule names

give explicit example of how reorder works

highlight that at least one rule needs to be specified

## 🚀 Links

- [Feature branch](https://developers.labs.livechat.com/docs/feature/API-8565-update_auto_access/management/configuration-api/v3.3/#auto-access)
- [Jira](https://livechatinc.atlassian.net/browse/API-8565)

## ✅ Checklist (for API docs)

- Changelog
- API versions
- Web & RTM
- **Table:** Available methods/webhooks/pushes
- **Table:** Specifics

For more guidelines, see [Updating API docs](https://livechatinc.atlassian.net/wiki/spaces/PAT/pages/761529400/Updating+API+docs).

### Content

- New content
- Proofreading/content fixes
  
- New feature
- Bug fix
- Breaking change

### DIFF
--- a/content/management/configuration-api/v3.3/index.mdx
+++ b/content/management/configuration-api/v3.3/index.mdx
@@ -975,7 +975,7 @@ If `next_id` is an empty string, then auto access will be placed at the end of t
 
 In the following list: **A -> B -> C -> D -> null**, letters represent the `auto_access` data structures while arrows indicate the `next_id` relation. Therefore, A has `next_id` set to B, B has `next_id` set to C, and so on.
 
-An `update_auto_access` call with the `{"id": "C", "next_id": "B"}` payload would result in the **A -> C -> B -> D -> null** sequence, maintaining consistency.
+An `update_auto_access` call with the `{"id": "C", "next_id": "B"}` payload would result in reordering the sequence, which after the successful request, would look like this: **A -> C -> B -> D -> null**. Consistency is maintained by keeping the `next_id` references properly assigned in each of the elements after a single request.
 
 #### Response
 
--- a/content/management/configuration-api/v3.3/index.mdx
+++ b/content/management/configuration-api/v3.3/index.mdx
@@ -744,7 +744,7 @@ Creates an `auto access` data structure, which is a set of conditions for the tr
 
 | Parameter      | Required | Data type | Example         |
 | -------------- | -------- | --------- | --------------- |
-| `country`      | No       | `string`  | "United States  |
+| `country`      | No       | `string`  | "United States" |
 | `country_code` | No       | `string`  | "US"            |
 | `region`       | No       | `string`  | "California"    |
 | `city`         | No       | `string`  | "Mountain View" |
@@ -954,16 +954,16 @@ Updates an existing `auto access`.  Only specified fields are updated (overwritt
 
 **2)** `<match_object>` structure:
 
-| Parameter     | Required | Data type | Notes                                                                                                   |
-| ------------- | -------- | --------- | ------------------------------------------------------------------------------------------------------- |
-| `value`       | Yes      | `string`  | A value to match                                                                                        |
-| `exact_match` | No       | `bool`    | If set to `false`, a `match_object.value` will be matched as a substring of the customer URL or domain. |
+| Parameter     | Required | Data type | Notes                                                                                                                   |
+| ------------- | -------- | --------- | ----------------------------------------------------------------------------------------------------------------------- |
+| `value`       | Yes      | `string`  | A value to match                                                                                                        |
+| `exact_match` | No       | `bool`    | If set to `false`, a `match_object.value` will be matched as a substring of the customer URL or domain; default: false. |
 
 **3)** the `geolocation_object` structure.  The condition is fulfilled only when all of the specified fields match.  At least one field must be specified.
 
 | Parameter      | Required | Data type | Example         |
 | -------------- | -------- | --------- | --------------- |
-| `country`      | No       | `string`  | "United States  |
+| `country`      | No       | `string`  | "United States" |
 | `country_code` | No       | `string`  | "US"            |
 | `region`       | No       | `string`  | "California"    |
 | `city`         | No       | `string`  | "Mountain View" |
--- a/content/management/configuration-api/v3.3/index.mdx
+++ b/content/management/configuration-api/v3.3/index.mdx
@@ -717,31 +717,30 @@ Creates an `auto access` data structure, which is a set of conditions for the tr
 | Method URL     | `https://api.livechatinc.com/v3.3/configuration/action/add_auto_access`       |
 | Required scopes| `access_rules:rw`                                                             |
 
-
 | Parameter                              | Required     | Data type                    | Notes                                                                                                   |
 | -------------------------------------- | ------------ | ---------------------------- | ------------------------------------------------------------------------------------------------------- |
 | `access`                               | Yes          | `object`                     | Destination access                                                                                      |
 | `access.groups`                        | Yes          | `[]int`                      | Destination groups                                                                                      |
-| `conditions`__*__                     | Yes          | `object`                     | Conditions to check                                                                                     |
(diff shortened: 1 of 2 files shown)


content/management/configuration-api/v3.3/index.mdx
```diff
@@ -925,23 +927,58 @@ curl -X POST \
 
 <Text>
 
-### Reorder Auto Access
+### Update Auto Access
 
-Moves an existing `auto access` data structure, specified by `id`, before another one, specified by `next_id`. 
-If `next_id` is empty, then the auto access will be placed at the end of the list.
+Updates existing `auto access`.  Only specified fields are updated (overwritten), leaving unspecified fields unchanged.
```
ale-kacperczyk, line 932:
```suggestion
Updates an existing `auto access`.  Only specified fields are updated (overwritten), leaving unspecified fields unchanged.
```

ale-kacperczyk (commented):
+ Changelog is missing

content/management/configuration-api/v3.3/index.mdx
```diff
@@ -925,23 +924,58 @@ curl -X POST \
...
-| `id`                   | Yes          | `string`  | ID of the auto access to move                                                                                                            |
-| `next_id`              | No           | `string`  | ID of the auto access that should follow the moved auto access. Leave empty if the moved auto access should be the last one on the list. |
+**4)** Modifying `next_id` moves the auto access specified by `id` before the auto access that is specified by `next_id`.
+If `next_id` is an empty string, then auto access will be placed at the end of the list. If you don't intend to reorder the rule at all, don't include `next_id` in the request. To better understand it, analyze the example below.
+
+💡 **Example:**
+
+In the following list: **A -> B -> C -> D -> null**, letters represent the `auto_access` data structures while arrows indicate the `next_id` relation. Therefore, A has `next_id` set to B, B has `next_id` set to C, and so on.
+
+An `update_auto_access` call with the `{"id": "C", "next_id": "B"}` payload would result in the **A -> C -> B -> D -> null** sequence, maintaining consistency.
```
ale-kacperczyk, line 978: What do you mean by "maintaining consistency"?

lpmichalik: it's that the backend assures that list always makes sense, so in the case described above notice that 'next_id' has to be modified for 3 different element, but the it's not client's responsability

ale-kacperczyk: Now when you pointed that out it makes sense. I'm not sure if we need to specify that, though. It may cause some doubts, but the client doesn't need to do anything.

lpmichalik: I was trying to be as explicit as possible as it was proven to be non-obvious.  @walaszczykm any thoughts if the description is clear enough and we can drop those last words?

walaszczykm: @ale-kacperczyk as @lpmichalik mentioned this change comes from the real-life scenario between us where I had trouble with understanding the real behaviour of this mechanism. IMO everything that is done automatically and kind-off magically under the hood by the server is something that is not obvious for the reader (because how could it be) and because of that, it is worth describing. When you encounter the description for the "update" method it is pretty natural by having some experience with various APIs that you expect to update the specific entity fields with optionally some of them restricted. Automatic reassigning of the `next_id` references for the elements in the list as it is happening here is totally out of the scope that could be easily assumed for this operation by the reader and have a real impact on the whole data structure. For example thanks to this addition in the documentation I will immediately know that after calling `update_auto_access` I don't need to call it multiple times for the rest of the elements to maintain the consistency of the list by my self.

ale-kacperczyk: The last sentence convinced me :D Is this short note enough, though?

walaszczykm: yes it is enough IMO maybe I would only try to put the word "reorder" somewhere in this sentence to clarify what type of operation it is that is being made so it would be obvious
```suggestion
An `update_auto_access` call with the `{"id": "C", "next_id": "B"}` payload would result in reordering the sequence which after the successful request will look like this: **A -> C -> B -> D -> null**. It maintainins its consistency by keeping the `next_id` references properly assigned in each of the elements after single request.
```

ale-kacperczyk, line 978:
```suggestion
An `update_auto_access` call with the `{"id": "C", "next_id": "B"}` payload would result in reordering the sequence, which after the successful request, would look like this: **A -> C -> B -> D -> null**. Consistency is maintained by keeping the `next_id` references properly assigned in each of the elements after a single request.
```

ale-kacperczyk: I added minor proofreading

ale-kacperczyk (approved):

walaszczykm (approved):

content/management/configuration-api/v3.3/index.mdx
```diff
@@ -925,23 +924,58 @@ curl -X POST \
...
+| Parameter     | Required | Data type | Notes                                                                                                   |
+| ------------- | -------- | --------- | ------------------------------------------------------------------------------------------------------- |
+| `value`       | Yes      | `string`  | A value to match                                                                                        |
+| `exact_match` | No       | `bool`    | If set to `false`, a `match_object.value` will be matched as a substring of the customer URL or domain. |
+
+**3)** the `geolocation_object` structure.  The condition is fulfilled only when all of the specified fields match.  At least one field must be specified.
+
+| Parameter      | Required | Data type | Example         |
+| -------------- | -------- | --------- | --------------- |
+| `country`      | No       | `string`  | "United States  |
```
bkubiak, line 966: missing `"` at the end

content/management/configuration-api/v3.3/index.mdx
```diff
@@ -925,23 +924,58 @@ curl -X POST \
...
+| `next_id`<sup>**4**</sup>                      | No          | `string`                      | ID of an existing auto access. If `next_id` is an empty string, auto_access moves to the end of the list. |
+
+**1)** `conditions` must have at least one of `url`, `domain`, `geolocation` set.
+
+**2)** `<match_object>` structure:
+
+| Parameter     | Required | Data type | Notes                                                                                                   |
+| ------------- | -------- | --------- | ------------------------------------------------------------------------------------------------------- |
+| `value`       | Yes      | `string`  | A value to match                                                                                        |
+| `exact_match` | No       | `bool`    | If set to `false`, a `match_object.value` will be matched as a substring of the customer URL or domain. |
```
bkubiak, line 960: I would like to see the default explicitly

bkubiak (approved):

content/management/configuration-api/v3.3/index.mdx
```diff
@@ -925,23 +924,58 @@ curl -X POST \
...
-| `next_id`              | No           | `string`  | ID of the auto access that should follow the moved auto access. Leave empty if the moved auto access should be the last one on the list. |
+**4)** Modifying `next_id` moves the auto access specified by `id` before the auto access that is specified by `next_id`.
+If `next_id` is an empty string, then auto access will be placed at the end of the list. If you don't intend to reorder the rule at all, don't include `next_id` in the request. To better understand it, analyze the example below.
+
+💡 **Example:**
+
+In the following list: **A -> B -> C -> D -> null**, letters represent the `auto_access` data structures while arrows indicate the `next_id` relation. Therefore, A has `next_id` set to B, B has `next_id` set to C, and so on.
+
+An `update_auto_access` call with the `{"id": "C", "next_id": "B"}` payload would result in reordering the sequence, which after the successful request, would look like this: **A -> C -> B -> D -> null**. Consistency is maintained by keeping the `next_id` references properly assigned in each of the elements after a single request.
```
bkubiak, line 978: The example is very clear 👏🏼

PR #875, opened 2021-03-30
(DPS-2748) Checked getting-started content

arturfracala:
## 🚀 Links

- [Feature branch](https://developers.labs.livechat.com/docs/feature/DPS-2748)
- [Jira](https://livechatinc.atlassian.net/browse/DPS-2748)

- New feature


public/images/getting-started/app-guides/index.mdx
```diff
@@ -0,0 +1,26 @@
+---
```
jakubsikora, line 1: Why this is in images ?

src/pages/getting-started/app-guides/agent-app-widgets/index.mdx
```diff
@@ -0,0 +1,71 @@
+---
```
jakubsikora, line 1: Did you use `git mv` for this content. It looks like it was created, but should be just moved. Same for other files.

jakubsikora (changes requested):
Nice work. Few comments. Please make sure you use `git mv` instead of copy/paste for the mdx file. I would like the history of these files to be kept.

jakubsikora (approved):
Nice work!

PR #894, opened 2021-04-21
Fix example payload for login v3.3+

kogierman:
## 🚀 Links

- [Feature branch](https://developers.labs.livechat.com/docs/feature/fix-login-example)

## 📓 Description
Small tweak of example payload


ale-kacperczyk (approved):
🥇

PR #963, opened 2021-06-18
[CW-2556] Custom Identity Provider docs

BartoszGamza:
# ❗️ Feature branch

Remember to start your branch name with `feature/`, for example, `feature/DPS-2947`. This will generate a preview on labs and allow us to test your changes.

# 🚀 Links

- [Feature branch](https://developers.labs.livechat.com/docs/feature/CW-2556-custom-identity-provider-docs)
- [Jira](https://livechatinc.atlassian.net/browse/CW-2556)

# 📓 Description

Docs for Custom Identity Provider feature we implemented in ChatWidget.


content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,173 @@
...
+
+
+# Examples
+
+Here are our suggested implementations of Customer Identity Provider. The `fetchLiveChatToken()` method used in examples is used to acquire the [Chat Widget token](#chat-widget-token).
+
+## Without caching
+
+If you don't need to cache the token the fetching logic shall occur on every refresh of your application page. In such case the implementation can be very simple - we only have to ensure handling of `getToken` promise and simply resolve others.
```
klarzynskik, line 72:
```suggestion
If you don't need to cache the token, the fetching logic shall occur on every refresh of your application page. In such a case, the implementation can be very simple - we only have to ensure handling of `getToken` promise and simply resolve others.
```

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,173 @@
...
+
+```js
+ window.__lc.custom_identity_provider = () => {
+    let tokenPromise = null
+    const fetchToken = () => {
+        tokenPromise = fetchLiveChatToken()
```
klarzynskik, line 78: maybe the `fetchLiveChatToken` could be also explained here, even with the pseudo-code? like, defined above and filled with comments, e.g.:

```js
const fetchLiveChatToken = function() {
        /* fetch a token from LiveChat Accounts API here */
}
```

but a bit more descriptive?

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,173 @@
...
+        getFreshToken: () => Promise.resolve(),
+        hasToken: () => Promise.resolve(false),
+        invalidate: () => Promise.resolve()
+    }
+ }
+```
+
+## With caching
+
+If you care about amount of requests being made in your application, you can unleash the full potential of Custom Identity Provider.
```
klarzynskik, line 96:
```suggestion
If you care about the number of requests being made in your application, you can unleash the full potential of the Custom Identity Provider.
```

> unleash the full potential

❤️

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,173 @@
...
+        getFreshToken,
+		hasToken,
+		invalidate,
+    }
+ }
+```
+
+## Repository
+
+Still hungry for knowledge? We have also prepared the example application presenting the usage of Custom Identity Provider mechanism. You can check it out [here](https://github.com/livechat/identity-provider-example/).
```
klarzynskik, line 172:
```suggestion
Still hungry for knowledge? We have also prepared the example application presenting the usage of the Custom Identity Provider mechanism. You can check it out [here](https://github.com/livechat/identity-provider-example/).
```

klarzynskik (approved):
Nicely done! 🚀 Left some minor comments regarding the grammar, and tiny suggestions.

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,173 @@
...
+# Chat Widget Token
+
+LiveChat authorization token you are willing to resolve in functions should contain the following properties:
+
+| Parameter      | Description                                                                      |
+| -------------- | -------------------------------------------------------------------------------- |
+| `accessToken`  | A token you can use to call LiveChat APIs on behalf of the user (customer).      |
+| `entityId`     | The ID of customer.                                                              |
+| `expiresIn`    | A number in **milliseconds** specifying how long the `accessToken` will be valid.|
+| `tokenType`    | Value: `Bearer`.                                                                 |
```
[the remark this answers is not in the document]
walaszczykm: @klarzynskik in case of a shape that describes the token object in the Chat Widget then yes it will always be an Bearer tokenType, this shape is not describing a direct output from accounts.

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,173 @@
...
+    window.__lc = window.__lc || {};
+    window.__lc.license = <LICENSE_NUMBER>;
+    window.__lc.custom_identity_provider = function() {
+      return {
+          getToken,
+          getFreshToken,
+          hasToken,
+          invalidate
+      }
+    }
```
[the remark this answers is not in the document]
BartoszGamza: it will throw because those functions are not defined :trollface: which should be pretty obvious, its described later - it would be an overkill to code everything here and then agin in examples imo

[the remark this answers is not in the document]
klarzynskik: > it will throw because those functions are not defined

yes, that's exactly what I'm referring to - it's described **later**, without any reference here what this actually does

we know that, as we did that, but IMO it's a bit illogical flow now, where the "mystery" is solved only later on

I'm not saying about coding that fully here, but perhaps make this more explicit, why these are not defined, or just rephrase that section?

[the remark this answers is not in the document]
klarzynskik: although it says "setup", it doesn't actually say anything about the setup itself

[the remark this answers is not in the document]
klarzynskik: maybe I'm a bit biased, but I don't think ppl just go through the docs as a whole 🤷🏼‍♂️ 

I'd rather say they go step-by-step, or even just scan for the interesting bits and pieces, and follow that - here, although it looks like something already cooked to be ready, is certainly not, with a lot of gaps to fill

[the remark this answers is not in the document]
BartoszGamza: Right, I could switch the order here. My aim to start from there was to imply where it should be put in the beginning but maybe its more logical to say what first and then define where

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,174 @@
...
+
+	const invalidate = () => {
+		cachedToken = null
+		retrievingToken = null
+		return storage.removeItem(cacheKey)
+	}
+
+    return {
+		getToken,
+        getFreshToken,
```
walaszczykm, line 164:
```suggestion
        	getFreshToken,
```

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,174 @@
...
+| `creationDate` | Unix timestamp specyfing the time of creation of the token.                       |
+| `licenseId`    | LiveChat license ID.                                                             |
+
+## How to acquire it?
+
+In order to get the token you need to implement authorizing function making a call to a `/customer` endpoint from LiveChat Accounts API. To do so, you can follow [creating a new customer flow](https://developers.livechat.com/docs/authorization/authorizing-api-calls/#creating-a-new-customer) or have a look at our [example](https://github.com/livechat/identity-provider-example/blob/master/lib/get-customer-token.ts) using cookies. Mind the fact that the [data](https://developers.livechat.com/docs/authorization/authorizing-api-calls/#response-4) returned from this endpoint is not 1:1 with the token shape described above. You will have to add the `creationDate` and `licenseId` fields. Additionally, we expect `expiresIn` to be defined in milliseconds while the endpoint operates in seconds. Therefore you should multiply the provided value by `1000` before passing it further.
```
walaszczykm, line 56:
```suggestion
In order to get the token, you need to implement a function that follows authorization flow making a call to a `/customer` endpoint from LiveChat Accounts API. To do so, you can follow [creating a new customer flow](https://developers.livechat.com/docs/authorization/authorizing-api-calls/#creating-a-new-customer) or have a look at our [example](https://github.com/livechat/identity-provider-example/blob/master/lib/get-customer-token.ts) using cookies. Mind the fact that the [data](https://developers.livechat.com/docs/authorization/authorizing-api-calls/#response-4) returned from this endpoint is not 1:1 with the token shape described above. You will have to add the `creationDate` and `licenseId` fields. Additionally, we expect `expiresIn` to be defined in milliseconds while the endpoint operates in seconds. Therefore you should multiply the provided value by `1000` before passing it further.
```

walaszczykm (approved):
Very well done 👏

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,174 @@
...
+```js
+ window.__lc.custom_identity_provider = () => {
+    const fetchLiveChatToken = () => {
+        // fetch a token from LiveChat Accounts API here
+    }
+
+    let tokenPromise = null
+    const fetchToken = () => {
+        tokenPromise = fetchLiveChatToken()
+            .then(response => {
```
Andarist, line 76: I feel that it's important to provide a comprehensive example here. We should use a catch handler here, clear the `tokenPromise` and rethrow the error (at the very least).

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,174 @@
...
+                tokenPromise = null
+                return response
+            })
+    }
+
+    return {
+        getToken: () => tokenPromise || fetchToken(),
+        getFreshToken: () => Promise.resolve(),
```
Andarist, line 84: This, according to this very documentation, should resolve to a token.

content/extending-chat-widget/custom-identity-provider/index.mdx
Andarist: Which also raises the need to document errors. We should either expect them to match the ones we create in the `@livechat/customer-auth` or we should learn to adapt whatever can be thrown here to our needs. I feel like probably the first one is better as it sets clear expectations, both on the consumer's and provider's sides of the story.

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,174 @@
...
+
+If you don't need to cache the token, the fetching logic shall occur on every refresh of your application page. In such a case, the implementation can be very simple - we only have to ensure handling of `getToken` promise and simply resolve others.
+
+```js
+ window.__lc.custom_identity_provider = () => {
+    const fetchLiveChatToken = () => {
+        // fetch a token from LiveChat Accounts API here
+    }
+
+    let tokenPromise = null
```
Andarist, line 73: I'm not sure if this example needs the whole dance with `tokenPromise` - it's a minor optimization in the `@livechat/customer-auth`. I believe that our widget shouldn't ever call `getToken` more than once.

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,174 @@
...
+        // fetch a token from LiveChat Accounts API here
+    }
+
+    const cacheKey = 'YOUR_CACHE_KEY'
+    let tokenPromise = null
+    let cachedToken = null
+
+    const isExpired = ({ creationDate, expiresIn }) => Date.now() >= creationDate + expiresIn
+
+    let retrievingToken = window.sessionStorage.getItem(cacheKey).then(token => {
```
Andarist, line 107: In what kind of an application the user of this solution could keep the token in the storage? It feels much safer to not do that ever and only keep the token in memory - after getting it from the backend.

content/extending-chat-widget/custom-identity-provider/index.mdx
walaszczykm: I have used this method in the example app and probably @BartoszGamza have to get it from there. Chat Widget was actually calling `getToken` twice sometimes so I have added it but maybe it is not needed for a simple example.

content/extending-chat-widget/custom-identity-provider/index.mdx
walaszczykm: No, as it will not be used when the `hasToken` always returns `Promise<false>` which is made to actually rely only on `getToken` which always gets token directly from the server without restoring it. This is a "simple type" example I have also used it in the example app to not complicate things too much.

content/extending-chat-widget/custom-identity-provider/index.mdx
walaszczykm: good point I think we shouldn't overcomplicate it too much so rethrowing an error should be good enough 👍

content/extending-chat-widget/custom-identity-provider/index.mdx
walaszczykm: and why it would not be safe to store it? you can use it for example to handle page refreshes without the need for additional calls. You get expiration time in the response so you can check on your own if the token should still be valid and if no, get the fresh one. This is exactly how we are doing it and it is a pretty common practices.

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,174 @@
...
+tagline: "Pair your own customer identity with LiveChat"
+desc: "Cross-platform chats for your applications"
+versionGroup: "chat-widget"
+---
+
+# Introduction
+
+## What is it for?
+
+If you have a more advanced service, handling user sessions, you can pair LiveChat customer identity with your own. As a result, your customers will never lose chats they made with your support. What is more, if you also use LiveChat on other platfroms (e.g., in mobile applications), their conversations will sync, making the communication even more seamless and natural.
```
konradk, line 15: I would add information that this kind of integration requires running its own backend service and it requires some development skills. As an alternative, we should mention the possibility to pass customer email/name using JS API.

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,174 @@
...
+          getFreshToken: ...,
+          hasToken: ...,
+          invalidate: ...
+      }
+    }
+```
+
+# Methods
+
+The aforementioned function assigned to `custom_identity_provider` property should return a set of the following promises:
```
konradk, line 34: I would indicate specifically who has to do what, i.e. "to use this feature, you must correctly implement and provide a set of following functions":

content/extending-chat-widget/custom-identity-provider/index.mdx
konradk: We should mention that such "function" needs to be a backend service

content/extending-chat-widget/custom-identity-provider/index.mdx
BartoszGamza: What is the relation between identtiy provider and customer email actually? One handles authorization - pass of identity created by our backend the other is just metadata that can be filled by customer

content/extending-chat-widget/custom-identity-provider/index.mdx
BartoszGamza: I think that GH messed something here, its Indented properly 😄

content/extending-chat-widget/custom-identity-provider/index.mdx
BartoszGamza: `getFreshToken` is being called after `invalidate` in customerSDK. In Maciej example it worked fine resolving nothing as this simply didn't happen as `getToken` is making a call each time and customerSDK **does not store the token on its own in any way - it calls `getToken` before each request to API**. To be 100% correct both `getToken` and `getFreshToken` should point to the exact same function as their functionalities overlap for this scenario, WDYT?

content/extending-chat-widget/custom-identity-provider/index.mdx
BartoszGamza: as I mentioned in other place it indeed gets called each time we make a request from customerSDK, therefore this is more of necessity than optimisation

content/extending-chat-widget/custom-identity-provider/index.mdx
BartoszGamza: as @walaszczykm says its very common practice and well - we do so. This whole example is more advanced but provides mature approach to the problem. I've seen several authorization logic implementations and its not obvious for people how to handle concurrent promises (one from api and other from storage) so I went with showing the good example how can it be done. If someone is a pro, he/she will do it on its own as it will be understood, if someone is still learning - it will be really appreciated. Plus, it works for us and therefore guarantees it will work for clients so if someone messes with own approach we can just ¯\_(ツ)_/¯ and link this example

content/extending-chat-widget/custom-identity-provider/index.mdx
walaszczykm: @konradk It is not true, it does not need to be called server-side. The customer authorization flow based on cookies does not require excachnging the `code` with corresponding `client_secret` for the token as it is done in the flow based on agent credentials. It is 100% doable on the client-side and we are doing exactly this in the CW.

The flow is as simple as: call `POST: http://api.livechat.com/v3.3/customer/token` with mainly `client_id` and `license` in the paylod and grab the token from the response. Cookies are being automatically set from the response by the browser so they will be also sent together with the next request to refresh the already existing customer identity.

The reason why we are doing it server-side in the `identity-provider-example` app is that we want to bind both identities: the service User and LiveChat Customer, together and provide Chat Widget with the appropriate Customer token for currently signed in User.

walaszczykm: In other words, you can use it totally on client-side and for example, use customer API to extend the capabilities of stock CW without the need for dedicated backend service. It is actually a pretty great use case for this feature that it allows extending the CW without the need to write it from scratch using `cusotmer-sdk` to handle some simple extended cases.

content/extending-chat-widget/custom-identity-provider/index.mdx
walaszczykm: actually, after rereading our CW code I think we should provide the `getToken` and `getFreshToken` with the same function as @BartoszGamza have said. Apart from inside the `getToken` we are calling `getFreshToken` also as a response for authorization error and redux action `PREFETCH_TOKEN`.

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,180 @@
...
+tagline: "Pair your own customer identity with LiveChat"
+desc: "Cross-platform chats for your applications"
+versionGroup: "chat-widget"
+---
+
+# Introduction
+
+## What is it for?
+
+If you have a more advanced service, handling user sessions, you can pair LiveChat customer identity with your own. As a result, your customers will never lose chats they made with your support. What is more, if you also use LiveChat on other platfroms (e.g., in mobile applications), their conversations will sync, making the communication even more seamless and natural. Mind the fact that Customer Identity is not customer metadata like name or email. If you would like to pass those you are probably looking for [updating session variables](https://developers.labs.livechat.com/docs/feature/CW-2556-custom-identity-provider-docs/extending-chat-widget/javascript-api/#update-session-variables).
```
ale-kacperczyk, line 15:
```suggestion
**Customer Identity Provider** was created for those who build more advanced integrations that handle user sessions. It allows you to match customer identities in LiveChat with those stored in your service. As a result, your customers will never lose chats they had with your support. What is more, if you also use LiveChat on other platforms (for example, in mobile applications), customer conversations will sync, making the communication even more seamless and natural. Mind the fact that _customer identity_ is not customer metadata such as name or email. If you would like to pass one of those, you are probably looking for [updating session variables](https://developers.labs.livechat.com/docs/feature/CW-2556-custom-identity-provider-docs/extending-chat-widget/javascript-api/#update-session-variables).
```

ale-kacperczyk (commented):
The first round of comments, let me know what you think so far.

content/extending-chat-widget/custom-identity-provider/index.mdx
ale-kacperczyk: I also feel that at this point, we should briefly explain "customer identity" since we're saying that it's not email or name

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,180 @@
...
+          getFreshToken: ...,
+          hasToken: ...,
+          invalidate: ...
+      }
+    }
+```
+
+# Methods
+
+In order to make it work, you'll have to properly implement and provide a set of following methods:
```
ale-kacperczyk, line 34:
```suggestion
In order to make Custom Identity Provider work, you'll have to properly implement and provide the following methods:
```

ale-kacperczyk: It's better to avoid ambiguous "it"

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+- `getToken` - resolving [Chat Widget token](#chat-widget-token). If you want to cache the token somehow, this should return the cached token instead of a fresh request to API.
```
ale-kacperczyk, line 36:
```suggestion
- `getToken` - resolving [Chat Widget token](#chat-widget-token). If you want to cache the token, this should return the cached token instead of a fresh request to API.
```

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,180 @@
...
+- `getFreshToken` - resolving [Chat Widget token](#chat-widget-token). This should always make a call for a fresh token from accounts API.
```
ale-kacperczyk, line 37:
```suggestion
- `getFreshToken` - resolving [Chat Widget token](#chat-widget-token). This should always make a call for a fresh token from the Accounts API.
```

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,180 @@
...
+- `hasToken` - resolving boolean. Determining whether a token has been acquired.
```
ale-kacperczyk, line 38:
```suggestion
- `hasToken` - resolving boolean. It determines whether a token has been acquired.
```

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,180 @@
...
+- `invalidate` - resolving nothing. When called should remove the current token. There is no need to do anything else, as a new token will be requested by `getFreshToken` afterward.
```
ale-kacperczyk, line 39:
```suggestion
- `invalidate` - resolving nothing. When called, it should remove the current token. There is no need to do anything else as a new token will be requested by `getFreshToken` afterward.
```

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+All of them will have to be returned in a function assigned to `window.__lc.custom_identity_provider`.
```
ale-kacperczyk, line 41: "them" should be replaced with a more specific word

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+# Chat Widget Token
+
+LiveChat authorization token you are willing to resolve in functions should contain the following properties:
```
ale-kacperczyk, line 45:
```suggestion
Here are the properties the LiveChat authorization token you want to resolve in functions should contain:
```

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,180 @@
...
+In order to make it work, you'll have to properly implement and provide a set of following methods:
+
+- `getToken` - resolving [Chat Widget token](#chat-widget-token). If you want to cache the token somehow, this should return the cached token instead of a fresh request to API.
+- `getFreshToken` - resolving [Chat Widget token](#chat-widget-token). This should always make a call for a fresh token from accounts API.
+- `hasToken` - resolving boolean. Determining whether a token has been acquired.
+- `invalidate` - resolving nothing. When called should remove the current token. There is no need to do anything else, as a new token will be requested by `getFreshToken` afterward.
+
+All of them will have to be returned in a function assigned to `window.__lc.custom_identity_provider`.
+
+# Chat Widget Token
```
ale-kacperczyk, line 43:
```suggestion
# Chat Widget token
```

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,180 @@
...
+| `creationDate` | Unix timestamp specyfing the time of creation of the token.                       |
+| `licenseId`    | LiveChat license ID.                                                             |
+
+## How to acquire it?
+
+In order to get the token, you need to implement a function that follows authorization flow by making a call to a `/customer` endpoint from LiveChat Accounts API. To do so, you can follow [creating a new customer flow](https://developers.livechat.com/docs/authorization/authorizing-api-calls/#creating-a-new-customer) or have a look at our [example](https://github.com/livechat/identity-provider-example/blob/master/lib/get-customer-token.ts) using cookies. Mind the fact that the [data](https://developers.livechat.com/docs/authorization/authorizing-api-calls/#response-4) returned from this endpoint is not 1:1 with the token shape described above. You will have to add the `creationDate` and `licenseId` fields. Additionally, we expect `expiresIn` to be defined in milliseconds while the endpoint operates in seconds. Therefore you should multiply the provided value by `1000` before passing it further.
```
ale-kacperczyk, line 58:
```suggestion
In order to get a Chat Widget token, you need to implement a function that calls the `/customer` endpoint. To do so, you can [create a new customer](https://developers.livechat.com/docs/authorization/authorizing-api-calls/#creating-a-new-customer) or have a look at our [example](https://github.com/livechat/identity-provider-example/blob/master/lib/get-customer-token.ts) that uses cookies. Mind the fact that the [data](https://developers.livechat.com/docs/authorization/authorizing-api-calls/#response-4) returned from the `enter-endpoint-here` endpoint is not 1:1 with the token shape described above. You'll have to make the following adjustments to your token before passing it further:

- Add the `creationDate` and `licenseId` fields. 
- Multiply `expiresIn` by 1000. It's because we expect `expiresIn` to be expressed in milliseconds while the `enter-endpoint-here` endpoint operates in seconds.
```

ale-kacperczyk: A few comments:
1. Can we cut out "follow authorizaton flow by making a call"? It seems to be unnecessarily lengthy.
2. The LiveChat Accounts API is another name to Global Accounts API. I would avoid using "LiveChat Accounts" and just paste full endpoints to avoid confusion.

content/extending-chat-widget/custom-identity-provider/index.mdx
ale-kacperczyk: I feel it would be better to enter the full endpoint here instead of calling it Accounts API/LiveChat Accounts API/ LiveChat Accounts because those are currently all synonyms of Global Accounts.

content/extending-chat-widget/custom-identity-provider/index.mdx
BartoszGamza: I agree on the `it` part, however when it comes to `a set of` I believe it describes better what it is meant to be - a JavaScript object rather than a list or array

content/extending-chat-widget/custom-identity-provider/index.mdx
BartoszGamza: "the above" would be enough?

content/extending-chat-widget/custom-identity-provider/index.mdx
BartoszGamza: I managed to describe it like: " _Customer identity_ is an id thats assigned to each customer by our API. Basing on that we can match data and chat history with certain person using our services." but Im not sure where to put it. Probably it should be defined at very beginning, maybe even above this paragraph? WDYT?

content/extending-chat-widget/custom-identity-provider/index.mdx
BartoszGamza: `you can [create a new customer]` is not what we want to achieve, essentially we would like to just get token of rather exisiting customer which is defined in the docs under the link - if we do have entity_id. Its all really tricky to describe without proper documentation of this use-case on the accounts side

BartoszGamza: probably it could be described a bit better but with following the flow, I just wanted the reader to study whats under this link and realise that they should have entity_id therefore its not really creating new customer but getting identity of already existing one. Still if they don't have a pair between their customer and ours (because it does not exist yet) they will have to create it so the name is valid only for the first execution of this logic

BartoszGamza:
Hi @ale-kacperczyk, thanks for your review! I've applied your suggestions and answered some comments. Would you mind taking a look again? 👀

content/extending-chat-widget/custom-identity-provider/index.mdx
BartoszGamza: I've tried to describe this better in recent commit

content/extending-chat-widget/custom-identity-provider/index.mdx
ale-kacperczyk: Okay 👍🏻

content/extending-chat-widget/custom-identity-provider/index.mdx
ale-kacperczyk: How about this one?

content/extending-chat-widget/custom-identity-provider/index.mdx
ale-kacperczyk: Better :)

content/extending-chat-widget/custom-identity-provider/index.mdx
ale-kacperczyk: Please, tak a look at this PR: https://github.com/livechat/livechat-public-docs/pull/964
Those are changes for Customer SSO, which will be released today. We got rid of the misleading "create a customer". You can link to the docs as it will be on prod before we merge your PR.

content/extending-chat-widget/custom-identity-provider/index.mdx
BartoszGamza: I didn't notice this suggestion at first butI feel like it lacks something (sounds really odd to me)

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,183 @@
...
+
+**Customer Identity Provider** was created for those who build more advanced integrations that handle user sessions. It allows you to match customer identities in LiveChat with those stored in your service. As a result, your customers will never lose chats they had with your support. What is more, if you also use LiveChat on other platforms (for example, in mobile applications), customer conversations will sync, making the communication even more seamless and natural. Mind the fact that _customer identity_ is not customer metadata such as name or email. If you would like to pass one of those, you are probably looking for [updating session variables](/extending-chat-widget/javascript-api/#update-session-variables).
+
+## Setup
+Custom Identity Provider is essentially a singleton function. You can define it as a part of the `LiveChatWidget` object from the tracking code used to install ChatWidget on your website or any other place accessing `window.__lc` object. However, **it has to be provided before widget initialization**.
```
ale-kacperczyk, line 20:
```suggestion

Custom Identity Provider is essentially a singleton function. You can define it as a part of the `LiveChatWidget` object from the tracking code used to install the Chat Widget on your website or any other place accessing `window.__lc` object. However, **it has to be provided before widget initialization**.
```

ale-kacperczyk: I' not sure about "...or any other place accessing `window.__lc` object"
Do you mean that one can access window.__lc anywhere on the page that has the tracking code? 

My problem is that "any other place" is quite vague.

content/extending-chat-widget/custom-identity-provider/index.mdx
ale-kacperczyk: I like the current description, though :) One remark: in the updates version of Customer SSO (customer auth docs) we call it "customer identity" not "new customer".

content/extending-chat-widget/custom-identity-provider/index.mdx
BartoszGamza: > Do you mean that one can access window.__lc anywhere on the page that has the tracking code?
yes

BartoszGamza: Its quite important to emphasise imo. Especially if someone uses more advanced setup than plain html page (like react or vue app etc.)

ale-kacperczyk: I also think we should keep it; I'm just thinking about a good way to rephrase it.

content/extending-chat-widget/custom-identity-provider/index.mdx
ale-kacperczyk: > The LiveChat authorization token you want to resolve in functions should contain the following properties:

How do you like this version? ☝🏻

content/extending-chat-widget/custom-identity-provider/index.mdx
ale-kacperczyk: I'm thinking about a bullet list. Let me know what you think.

Custom Identity Provider is essentially a singleton function. You can define it as part of one of the following objects: 
- the `LiveChatWidget` object from the LiveChat tracking code
- the `window.__lc` object you can access anywhere on a page that has the LiveChat tracking code

In both cases, **it has to be provided before widget initialization**.

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,183 @@
...
+          invalidate: ...
+      }
+    }
+```
+
+# Methods
+
+In order to make Custom Identity Provider work, you'll have to properly implement and provide a set of following methods:
+
+- `getToken` - resolving [Chat Widget token](#chat-widget-token). If you want to cache the token, this should return the cached token instead of a fresh request to API.
```
ale-kacperczyk, line 38: I missed this one. 
"...fresh request to API". We should specify which API or maybe even better, the endpoint

content/extending-chat-widget/custom-identity-provider/index.mdx
BartoszGamza: I've added: "from `https://accounts.livechat.com/customer/` endpoint."

content/extending-chat-widget/custom-identity-provider/index.mdx
BartoszGamza: they are both the same thing actually, its called LiveChatWidget object in https://developers.livechat.com/docs/extending-chat-widget/javascript-api/#getting-started but its assigned to window.__lc property. By splitting it Im afraid we will intorduce some difference that doesn't exist. How about making `LiveChatWidget object` a link to the one I provided here?

ale-kacperczyk: No splitting, then :)

> You can define it as part of the `LiveChatWidget` object from the LiveChat tracking code used on your website or from any other place that has access to the `window._lc` object (might be the case in more advanced applications).

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,183 @@
...
+| `licenseId`    | LiveChat license ID.                                                             |
+
+## How to acquire it?
+
+In order to get a Chat Widget token, you need to implement a function that calls the `https://accounts.livechat.com/customer/token` endpoint. To do so, you can follow [customer authorization flows](https://developers.livechat.com/docs/authorization/authorizing-api-calls/#customer-authorization-flows) or have a look at our [example](https://github.com/livechat/identity-provider-example/blob/master/lib/get-customer-token.ts) that uses cookies. For the first time, if you do not have pair between our `customer_id` (`entity_id`) and identity in your application, you will be creating new customer identity. For any further execution of this logic you will be able to provide `entity_id` and therefore you will be just refreshing the token. Mind the fact that the [data](https://developers.livechat.com/docs/authorization/authorizing-api-calls/#response-4) returned from the endpoint is not 1:1 with the token shape described above. You'll have to make the following adjustments to your token before passing it further:
```
ale-kacperczyk, line 60:
```suggestion
In order to get a Chat Widget token, you need to implement a function that calls the `https://accounts.livechat.com/customer/token` endpoint. To do so, you can follow [customer authorization flows](https://developers.livechat.com/docs/authorization/authorizing-api-calls/#customer-authorization-flows) or have a look at our [example](https://github.com/livechat/identity-provider-example/blob/master/lib/get-customer-token.ts) that uses cookies. 

The first time you execute this logic, the connection between our `customer_id` (`entity_id`) and the identity from your service doesn't exist yet. Start by creating a new customer identity. Next time you execute this logic, you'll be able to provide `entity_id`, and therefore, you will only need to refresh the token. Mind the fact that the [data](https://developers.livechat.com/docs/authorization/authorizing-api-calls/#response-4) returned from the endpoint is not 1:1 with the token shape described above. You'll have to make the following adjustments to your token before passing it further:
```

ale-kacperczyk: I'd separate these two paragraphs for better clarity.

content/extending-chat-widget/custom-identity-provider/index.mdx
```diff
@@ -0,0 +1,183 @@
...
+        getFreshToken,
+		hasToken,
+		invalidate,
+    }
+ }
+```
+
+## Repository
+
+Still hungry for knowledge? We have also prepared the example application presenting the usage of the Custom Identity Provider mechanism. You can check it out [here](https://github.com/livechat/identity-provider-example/).
```
ale-kacperczyk, line 182:
```suggestion
Still hungry for knowledge? We have also prepared an example application presenting the usage of the Custom Identity Provider mechanism. You can check it out [here](https://github.com/livechat/identity-provider-example/).
```

ale-kacperczyk (approved):

PR #981, opened 2021-07-12
(DPS-3084) Revamp Management Overview

ale-kacperczyk:
# ❗️ Feature branch

Remember to start your branch name with `feature/`, for example, `feature/DPS-2947`. This will generate a preview on labs and allow us to test your changes.

# 🚀 Links

- [Feature branch](https://developers.labs.livechat.com/docs/feature/DPS-3084)
- [Jira](https://livechatinc.atlassian.net/browse/DPS-3084)

# 📓 Description
Currently, **Messaging Overview** contains a [section about the Configuration API](https://developers.livechat.com/docs/messaging/#configuration-api), which doesn't belong there. I moved the section to Management Overview and extended it.

### DIFF
--- a/content/getting-started/livechat-apps/webhook-apps/index.mdx
+++ b/content/getting-started/livechat-apps/webhook-apps/index.mdx
@@ -97,6 +97,9 @@ The settings page will appear right after a LiveChat user installs your app from
 Once you have the LiveChat credentials, you can register your webhooks with the Configuration API. Currently, there are two APIs you can use to manage webhooks:
 
 - [Configuration API v3](/management/configuration-api/#webhooks)
+  
+  💡 Read the tutorial on  <a href="https://developers.livechat.com/updates/setting-up-webhooks-with-livechat-api" target="_blank">how to register LiveChat webhooks v3.</a>
+
 - [Configuration API v2](/management/configuration-api/v2.0/#create-a-new-webhook) (deprecated)
 
 ### Setting up webhooks v2 for development
--- a/content/management/index.mdx
+++ b/content/management/index.mdx
@@ -8,54 +8,127 @@ desc: "There are dedicated APIs for the resource management. Use the Configurati
 
 # Introduction
 
-LiveChat Platform provides multiple APIs for managing resources.
-
-# Configuration API
+The Configuration API allows you to manage resources such as **agents, auto access, bots, groups, properties, and webhooks**. You can use it to build advanced integrations as well as alter the configuration of your resources directly via the API.
 
 ## Use cases
 
-The [**Configuration API**](/management/configuration-api/) allows for:
+Here are some of the most common use cases of the Configuration API:
 
-- storing license configurations
-- creating chat, thread, and event [properties](/management/configuration-api/#properties)
-- managing [webhooks](/management/configuration-api/#webhooks), for example registering and unregistering
-- managing [Bots](/management/configuration-api/#bots), for example creating and removing
+- Build integrations that connect LiveChat with other services via **webhooks**. React to events happening in the chat.
+- Create automation solutions with the use of **bots**. Connect LiveChat with various AI solutions.
+- Use **properties** to store information about chats, threads, or events.
 
-In the near future, it will allow for groups configuration. For now, refer to [**Configuration API v2**](/management/configuration-api/v2.0/#groups) to manage groups properties.
+<!-- In the near future, it will allow for groups configuration. For now, refer to [**Configuration API v2**](/management/configuration-api/v2.0/#groups) to manage groups properties. -->
 
-## Properties
+## API versioning
 
-Properties are **key-value storages**. They can be set within a chat, a thread, or an event. You can create properties within a license and configure them using the [Configuration API](/management/configuration-api/). It's possible to configure the property [type](../management/configuration-api/#property-types), [location](../management/configuration-api/#property-locations), and [domain](../management/configuration-api/#property-domain).
+The LiveChat Configuration API adopts the following versioning convention:
 
-Refer to the [Configuration API](/management/configuration-api/) document to read more about the [property format](/management/configuration-api/#properties) and available [endpoints](/management/configuration-api/#methods-1).
+- **stable**
 
-## Webhooks
+  The latest stable version we encourage developers to use. You can only expect bug fixes; no new new functionalities or breaking changes will be introduced.
 
-In general, actions invoked by the use of Messaging APIs result in events. You can be notified about those events with **webhooks**. LiveChat provides a number of webhooks, which you can register and manage via the [Configuration API](../management/configuration-api/#webhooks).
-We distinguish **global webhooks** and **bot-specific webhooks**. Once **global webhooks** are set up, you will always receive them. **Bot-specific** webhooks are strongly coupled with the bot's status (`accepting chats`, `not accepting chats`, `offline`). If the bot is offline, webhooks won't be received.
-Webhook equivalents in the websocket transport (RTM API) are **pushes**.
+- **developer preview (dev preview)**
+
+  The unstable version that provides a preview of the upcoming functionalities. **It's not open for public use by default.** If you want to use it, contact us at developers@livechat.com or <a href="#open-chat" onClick={openChatWindow}>ask on the chat</a>, and we'll give you access.
+
+- **legacy**
+
+  The old version we still support. We highly encourage you to switch to the latest stable version.
+
+**Currently:**
+
+|                 | Configuration API |
+| --------------- | ----------------- |
+| **stable**      | 3.3               |
+| **dev preview** | 3.4               |
+| **legacy**      | 2.0, 3.1, 3.2     |
+
+<Warning>
+
+Currently, Configuration API v2 and v3 cover different resources. However, if you're still using v2, we encourage you to switch to v3 as more and more resources are being migrated to v3.
+
+</Warning>
+
+# Resources
+
+## Agents
+
+An agent is a type of user who communicates with customers. You can see the agent data structure [here](/messaging/agent-chat-api/#agent). The Configuration API allows for multiple agent-related actions, including creating, updating, deleting, and suspending agents.
+
+<SectionLink to={"/management/configuration-api/#agents"}>
+  See agent methods
+</SectionLink>
+
+## Auto access
+
+Auto access is a functionality that allows for assigning customers to dedicated groups based on the URL or the customer's geolocation.
+
+<SectionLink to={"/management/configuration-api/#auto-access"}>
+  See auto access methods
+</SectionLink>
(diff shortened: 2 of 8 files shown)


content/messaging/index.mdx
```diff
@@ -116,24 +121,24 @@ If you're not sure, which implementation to choose, we suggest reading about par
...
   An unstable version that provides a preview of the upcoming features. **It's not open to public use.** However, if you want to test some features, contact us at developers@livechat.com or <a href="#open-chat" onClick={openChatWindow}>ask on the chat</a>, and we'll give you access.
 
 - **legacy**
-  
+
   The old version we still support but highly encourage you to switch to the latest stable version.
 
 **Currently:**
 
-|                 | Agent Chat API | Customer Chat API | Configuration API |
```
ale-kacperczyk, line 132: I removed Config API from this table to be consistent. I think we should add a similar table to Reports API.

content/messaging/index.mdx
```diff
@@ -102,12 +102,12 @@ Agent Chat API and Customer Chat API can be used either as **Real-Time Messaging
 
 To learn about differences between these two API types, see the comparison below:
 
-|          **Characteristics**          |                                                                         **RTM API**                                                                         |                                                           **Web API**                                                           |
```
ale-kacperczyk, line 105: This is just formatting. Sorry

content/management/index.mdx
```diff
@@ -8,20 +8,125 @@ desc: "There are dedicated APIs for the resource management. Use the Configurati
...
+| **dev preview** | 3.4               |
+| **legacy**      | 2.0, 3.1, 3.2     |
+
+<Warning>
+
+Currently, Configuration API v2 and v3 cover different resources. However, if you're still using v2, we encourage you to switch to v3 as more and more resources are being migrated to v3.
+
+</Warning>
+
+# Resources
```
ale-kacperczyk, line 51: I'm not sure about this headline as we tend to use "data structures" instead of "resources". But "data structures" seems flat to me... As if it was just a JSON. WDYT?

bkubiak: I like "resources" here, when we write "data structures" we refer only to the shape of some objects, not the resource itself.

bkubiak: I like "resources" here, when we write "data structures" we refer only to the shape of some objects, not the resource itself.

content/management/index.mdx
```diff
@@ -8,20 +8,125 @@ desc: "There are dedicated APIs for the resource management. Use the Configurati
...
+
+## Agents
+
+An agent is a type of user who communicates with customers. You can see the agent data structure [here](/messaging/agent-chat-api/#agent). The Configuration API allows for multiple agent-related actions, including creating, updating, deleting, and suspending agents.
```
bkubiak, line 55: >You can see the agent data structure...
...how it looks as a user in a chat.

Anyway, I'm not sure if linking to the agent **in chat** is ok, because it's configuration-api and agent data structure here looks a bit different (see `get-agent` method).

content/management/index.mdx
```diff
@@ -8,20 +8,125 @@ desc: "There are dedicated APIs for the resource management. Use the Configurati
...
+
+<SectionLink to={"/management/configuration-api/#agents"}>
+  See agent methods
+</SectionLink>
+
+## Auto access
+
+Auto access is a functionality that allows for assigning customers to dedicated groups based on the URL or the customer's geolocation.
```
bkubiak, line 63: >to dedicated groups based on
based on different rules like a URL of the website or the customer's geolocation.

In fact, there are more rules to set, but we don't need to mention all of them here.

How about describing full context here? Because a developer has no idea what is customer group and why they would want to set it (automatically).

>...When such customer starts a chat, the chat will be routed to the specific group.

content/management/index.mdx
```diff
@@ -8,20 +8,125 @@ desc: "There are dedicated APIs for the resource management. Use the Configurati
...
+## Webhooks
+
+In a nutshell, actions invoked by the use of Messaging APIs or the Configuration API result in events. You can be notified about those events with webhooks. LiveChat provides a number of webhooks, which you can register and manage via the Configuration API. Just like properties, webhooks are registered per Client Id (integration), not per license (v3.3+).
+
+We can distinguish two types of webhooks:
+
+- license webhooks
+- bot webhooks
+
+Once **license webhooks** are set up, you will always receive them. Read our tutorial on <a href="https://developers.livechat.com/updates/setting-up-webhooks-with-livechat-api/" target="_blank">how to configure LiveChat license webhooks via the Configuration API</a>.
```
bkubiak, line 110: by setting up you meain registering and **enabling** them, right?

content/management/index.mdx
```diff
@@ -8,20 +8,125 @@ desc: "There are dedicated APIs for the resource management. Use the Configurati
...
+  See auto access methods
+</SectionLink>
+
+## Bots
+
+Bots are similar to regular agents. The main difference between them is that bots have a greater potental for automation, for example, they can listen for and react to incoming webhooks and pushes.
+
+Bots are created and managed via the Configuration API. Just as regualar agents, bots can call Agent Chat API by the Web or RTM API.
+
+Bots are authorized with the use of the [agent token](/authorization/authorizing-api-calls/#agent-authorization-flows). Using bots requires sending the `X-Author-Id` header for the Web API and the `author_id` property for the RTM API.
```
bkubiak, line 75: Perhaps, we should mention here that bot status can be set by using agent-chat-api and `set_routing_status` method. It can be important here.

bkubiak (commented):
Good changes, I like it! It will be more clear now what configuration API is.

content/management/index.mdx
ale-kacperczyk: We can link to Get Agents 👍🏻 And we can also change it here: https://developers.livechat.com/docs/management/configuration-api/#agents

content/management/index.mdx
ale-kacperczyk: That might be a good idea

content/management/index.mdx
ale-kacperczyk: Yeah :) I think it's okay for a general overview because a developer might want to "do something with properties, some configuration/setup), and doesn't really know that props need to be registered & enabled.

content/management/index.mdx
ale-kacperczyk: We can do that 👍🏻

content/getting-started/livechat-apps/webhook-apps/index.mdx
```diff
@@ -97,6 +97,9 @@ The settings page will appear right after a LiveChat user installs your app from
 Once you have the LiveChat credentials, you can register your webhooks with the Configuration API. Currently, there are two APIs you can use to manage webhooks:
 
 - [Configuration API v3](/management/configuration-api/#webhooks)
+
```
ale-kacperczyk, line 100: I also added the link you suggested via DM

content/management/index.mdx
bkubiak: ok

bkubiak (approved):

content/management/index.mdx
```diff
@@ -8,20 +8,127 @@ desc: "There are dedicated APIs for the resource management. Use the Configurati
...
+
+An agent is a type of user who communicates with customers. You can see the sample agent data structure in the [response of Get Agent](/management/configuration-api/#get-agent). The Configuration API allows for multiple agent-related actions, including creating, updating, deleting, and suspending agents.
+
+<SectionLink to={"/management/configuration-api/#agents"}>
+  See agent methods
+</SectionLink>
+
+## Auto access
+
+Auto access is a functionality that allows for assigning customers to dedicated groups based on the URL or the customer's geolocation. When a customer who matches some pre-defined auto access rules starts a chat, the chat is automatically routed to the specific group. You can use auto access, for example, to provide multi-language support with each group of agents being responsible for communication in a different language.
```
oliwiapolec, line 63:
```suggestion
Auto access is a functionality that allows for assigning customers to dedicated groups based on the visited URL or the customer's geolocation. When a customer who matches some pre-defined auto access rules starts a chat, the chat is automatically routed to the specific group. You can use auto access, for example, to provide multi-language support with each group of agents being responsible for communication in a different language.
```

content/management/index.mdx
```diff
@@ -8,20 +8,127 @@ desc: "There are dedicated APIs for the resource management. Use the Configurati
...
+
+<SectionLink to={"/management/configuration-api/#auto-access"}>
+  See auto access methods
+</SectionLink>
+
+## Bots
+
+Bots are similar to regular agents. The main difference between them is that bots have a greater potental for automation, for example, they can listen for and react to incoming webhooks and pushes.
+
+Bots are created and managed via the Configuration API. Just as regualar agents, bots can call Agent Chat API by the Web or RTM API.
```
oliwiapolec, line 73:
```suggestion
Bots are created and managed via the Configuration API. Just as regular agents, bots can call the Agent Chat API by either Web or RTM API.
```

content/management/index.mdx
```diff
@@ -8,20 +8,127 @@ desc: "There are dedicated APIs for the resource management. Use the Configurati
...
+
+To change a bot's status (`accepting_chats`, `not_accepting_chats`, or `offline`), use the [Set Routing Status](/messaging/agent-chat-api/#set-routing-status) method from the Agent Chat API.
+
+<SectionLink to={"s/management/configuration-api/#bots"}>
+  See bot methods
+</SectionLink>
+
+## Groups
+
+Groups let you organize your work by creating teams with agent and bot members. You can use groups to do configuration based on shared settings (language, working hours, department) and to separate chat routing.
```
oliwiapolec, line 85: This one seems a little short compared to the other ones, maybe we can add a note similar to the one we added for the agents?

```suggestion
Groups let you organize your work by creating teams with agent and bot members. You can use groups to do configuration based on shared settings (language, working hours, department) and to separate chat routing. With the Configuration API, you can perform actions such as creating and deleting a group, or updating its' settings.
```

content/management/index.mdx
```diff
@@ -8,20 +8,127 @@ desc: "There are dedicated APIs for the resource management. Use the Configurati
...
 
-## New Configuration APIs
+## Use cases
 
-Configuration API is a service for storing configuration of license. It allows you to set up custom properties, bot agents and webhooks. You'll be using this API together with [Messaging APIs](../messaging/).
+Here are some of the most common use cases of the Configuration API:
 
-<SectionLink to={"/management/configuration-api/"}>
-  Configuration API Reference
+- Build integrations that connect LiveChat with other services via **webhooks**. React to events happening in the chat.
```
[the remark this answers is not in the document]
bkubiak: you're right

oliwiapolec (approved):

PR #992, opened 2021-07-29
[no-dps] Remove Limits section + misc fixes

oliwiapolec:
# 📓 Description

We need to temporarily delete the "Limits" section since the changes aren't on prod and the hidden document was searchable in docs search


ale-kacperczyk:
Solved in https://github.com/livechat/livechat-public-docs/pull/994. We needed to take a different approach and create a 302 redirect instead of removing the docs

PR #994, opened 2021-08-02
(NO-DPS) Add 302 redirect for Limits

ale-kacperczyk:
# 📓 Description

Instead of removing the content from https://developers.livechat.com/docs/management/limits, we'll add a temp redirect to the LiveChat pricing page.


jakubsikora (approved):
Lovely

PR #1073, opened 2021-12-13
(DPS-3399) The highlighted space around an item in sidebar isn't clickable

arturfracala:
# ❗️ Feature branch / Deploy Preview

We suggest to name your branch with `feature/`, for example, `feature/DPS-2947`. See below notifications from **Netlify** about the url.

# 🚀 Links

- [Jira](https://livechatinc.atlassian.net/browse/DPS-3399)

# 📓 Description

### DIFF
--- a/src/components/core/SideNav/MenuItem.js
+++ b/src/components/core/SideNav/MenuItem.js
@@ -9,7 +9,6 @@ const LinkWrapper = styled.div`
   display: flex;
   align-items: center;
   justify-content: space-between;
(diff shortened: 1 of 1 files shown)


dominikslawkowski (approved):

jakubsikora (approved):
Just a question. This is already working on the homepage sidebar in prod. Can't we reuse the same component and styling, instead of creating new styled component ?

arturfracala:
> Just a question. This is already working on the homepage sidebar in prod. Can't we reuse the same component and styling, instead of creating new styled component ?

More useful would be replacing homepage sidebar items with that component. That refactor would have more sense. I'll take this into account

PR #1084, opened 2021-12-28
Update vulnerable libs

jakubsikora:
Update vulnerable libs


jakubsikora:
We did the regression with this branch.

PR #1091, opened 2022-01-13
Setup code ownership rules

ale-kacperczyk:
# Description
Added CODEOWNERS file to protect `master`. We will require approval from a member of @livechat/tech-content  when any file in `payload` or `pages` is changed


jakubsikora (approved):
Thats very kind xd

PR #1092, opened 2022-01-17
Fix code sample in Customer Identity Provider

BartoszGamza:
# Deploy preview

Scroll down for the link to the deploy preview.

# Links

n/a
- [Jira](https://livechatinc.atlassian.net/browse/XXX-XXX)

# Description

@bolchowka noticed that the code example in [Customer Identity Provider](https://developers.livechat.com/docs/extending-chat-widget/custom-identity-provider#without-caching) (w/o caching) won't work because `fetchToken()` function does not return a promise - it only assigns one. The additional return should resolve the confusion 😉


bolchowka (approved):

klarzynskik (approved):

ale-kacperczyk (approved):
Thanks for the fix :)

PR #1126, opened 2022-04-08
(DED-87) Automatic deploys in docs

arturfracala:
# Links

- [Jira](https://livechatinc.atlassian.net/browse/DED-87)

# Description


.github/workflows/preview.yml
```diff
@@ -0,0 +1,13 @@
+name: Create preview
```
jakubsikora, line 1: Why you are not using same action library for previews ? I can see they support it https://github.com/jsmrcaga/action-netlify-deploy#use-branch-name-to-deploy even with generating the preview url in the checks section.

jakubsikora (changes requested):

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,19 @@
+name: Deploy branch
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  deployer:
+    runs-on: ubuntu-latest
```
jakubsikora, line 10: Please also setup timeout to avoid draining our spendings limits:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes
I suggest: `timeout-minutes: 30`

.github/workflows/preview.yml
```diff
@@ -0,0 +1,29 @@
...
+
+on:
+  push:
+    branches-ignore:
+      - master
+
+jobs:
+  previewCreator:
```
jakubsikora, line 9: Also timeout for this job.

jakubsikora (changes requested):

jakubsikora (approved):

PR #1145, opened 2022-05-05
Describe v3.5 tags.

gluszczyk:
# Links

- [Jira](https://livechatinc.atlassian.net/browse/API-10298)

PR #1146, opened 2022-05-06
DED-138 Tag thread and untag thread descriptions added

kamilatokarska:
# Deploy preview

# Links

Link your Jira ticket.

- [Jira]()

# Description

# Review and release

### DIFF
--- a/src/pages/messaging/agent-chat-api/index.mdx
+++ b/src/pages/messaging/agent-chat-api/index.mdx
@@ -1583,8 +1583,14 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
 
 ### Tag Thread
 
-Active and inactive chats can be tagged, and one or more tags can be added to the specific **thread**. To tag a chat, the **tag** needs to be created first. It's possible to do that in the [LiveChat Application](https://accounts.livechat.com/) or via the [Configuration API](/management/configuration-api). Tags are case sensitive and are defined by each client individually. The **list of tags** can be found in the [LiveChat Application](https://my.livechatinc.com/settings/tags). 
+Tags a given thread. The following restrictions apply:
 
+- Active and inactive chats can be tagged.
+- One or more tags can be added to the specific **thread**.
+- To tag a chat, the **tag** needs to be created first. It can be done in the [LiveChat Application](https://accounts.livechat.com/) or via the [Configuration API](/management/configuration-api).
+- Tags are case sensitive.
+- Tags are defined by each client individually. The **list of tags** can be found in the [LiveChat Application](https://my.livechatinc.com/settings/tags). 
+   
 The **tagged threads** can be untagged using [Untag Thread](#untag-thread).
 
 #### Specifics
@@ -1633,7 +1639,11 @@ https://api.livechatinc.com/v3.4/agent/action/tag_thread \
 
 ### Untag Thread
 
-To make a request, you need to specify the **tag name**, the **Id of the chat** and the **Id of the thread** you want to remove a **tag** from. Active and inactive chats can be untagged, and one or more tags can be removed from the specific **thread**. Tags are case sensitive.
+Untags a given thread. The following restrictions apply:
+
+- Active and inactive chats can be untagged. 
+- One or more tags can be removed from the specific **thread**. 
+- Tags are case sensitive.
 
 The **untagged threads** can be tagged using [Tag Thread](#tag-thread).
 
--- a/src/pages/messaging/agent-chat-api/index.mdx
+++ b/src/pages/messaging/agent-chat-api/index.mdx
@@ -1583,15 +1583,9 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
 
 ### Tag Thread
 
-Tags a given thread. The following restrictions apply:
+Tags a given **thread**. 
 
-- Active and inactive chats can be tagged.
-- One or more tags can be added to the specific **thread**.
-- To tag a chat, the **tag** needs to be created first. It can be done in the [LiveChat Application](https://accounts.livechat.com/) or via the [Configuration API](/management/configuration-api/v2.0#tags).
-- Tags are case sensitive.
-- Tags are defined by each client individually. The **list of tags** can be found in the [LiveChat Application](https://my.livechatinc.com/settings/tags). 
-   
-The **tagged threads** can be untagged using [Untag Thread](#untag-thread).
+[Read how to list all available chats](https://my.livechatinc.com/settings/tags) and [create a new tag](/management/configuration-api/v2.0#add-a-tag). **Tags** are case sensitive in this method.
 
 #### Specifics
 
@@ -1639,13 +1633,7 @@ https://api.livechatinc.com/v3.4/agent/action/tag_thread \
 
 ### Untag Thread
 
-Untags a given thread. The following restrictions apply:
-
-- Active and inactive chats can be untagged. 
-- One or more tags can be removed from the specific **thread**. 
-- Tags are case sensitive.
-
-The **untagged threads** can be tagged using [Tag Thread](#tag-thread).
+Untags a given **thread**. **Tags** are case sensitive in this method.
 
 #### Specifics
 
--- a/src/pages/messaging/agent-chat-api/index.mdx
+++ b/src/pages/messaging/agent-chat-api/index.mdx
@@ -1583,9 +1583,9 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
 
 ### Tag Thread
 
-To make a request, you need to specify the **tag name**, the **Id of the chat** and the **Id of the thread** you want to add a **tag** to. Active and inactive chats can be tagged, and one or more tags can be added to the specific **thread**. To tag a chat, the **tag** needs to be created first. It's possible to do that in the [LiveChat Application](https://accounts.livechat.com/) or via the [Configuration API](/management/configuration-api). Tags are case sensitive and are defined by each client individually. The **list of tags** can be found in the [LiveChat Application](https://my.livechatinc.com/settings/tags). 
+Active and inactive chats can be tagged, and one or more tags can be added to the specific **thread**. To tag a chat, the **tag** needs to be created first. It's possible to do that in the [LiveChat Application](https://accounts.livechat.com/) or via the [Configuration API](/management/configuration-api). Tags are case sensitive and are defined by each client individually. The **list of tags** can be found in the [LiveChat Application](https://my.livechatinc.com/settings/tags). 
 
-The **tagged threads** can be untagged using [Untag Thread](/messaging/agent-chat-api#untag-thread).
+The **tagged threads** can be untagged using [Untag Thread](#untag-thread).
 
 #### Specifics
 
@@ -1635,7 +1635,7 @@ https://api.livechatinc.com/v3.4/agent/action/tag_thread \
 
 To make a request, you need to specify the **tag name**, the **Id of the chat** and the **Id of the thread** you want to remove a **tag** from. Active and inactive chats can be untagged, and one or more tags can be removed from the specific **thread**. Tags are case sensitive.
 
-The **untagged threads** can be tagged using [Tag Thread](/messaging/agent-chat-api#tag-thread).
+The **untagged threads** can be tagged using [Tag Thread](#tag-thread).
 
 #### Specifics
 
--- a/src/pages/messaging/agent-chat-api/index.mdx
+++ b/src/pages/messaging/agent-chat-api/index.mdx
@@ -1583,6 +1583,10 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
 
 ### Tag Thread
 
+To make a request, you need to specify the **tag name**, the **Id of the chat** and the **Id of the thread** you want to add a **tag** to. Active and inactive chats can be tagged, and one or more tags can be added to the specific **thread**. To tag a chat, the **tag** needs to be created first. It's possible to do that in the [LiveChat Application](https://accounts.livechat.com/) or via the [Configuration API](/management/configuration-api). Tags are case sensitive and are defined by each client individually. The **list of tags** can be found in the [LiveChat Application](https://my.livechatinc.com/settings/tags). 
+
+The **tagged threads** can be untagged using [Untag Thread](/messaging/agent-chat-api#untag-thread).
+
 #### Specifics
 
 |                        |                                                                          |
@@ -1628,6 +1632,10 @@ https://api.livechatinc.com/v3.4/agent/action/tag_thread \
 <Text>
 
 ### Untag Thread
+
+To make a request, you need to specify the **tag name**, the **Id of the chat** and the **Id of the thread** you want to remove a **tag** from. Active and inactive chats can be untagged, and one or more tags can be removed from the specific **thread**. Tags are case sensitive.
+
+The **untagged threads** can be tagged using [Tag Thread](/messaging/agent-chat-api#tag-thread).
 
 #### Specifics
 
--- a/src/pages/messaging/agent-chat-api/index.mdx
+++ b/src/pages/messaging/agent-chat-api/index.mdx
@@ -1587,7 +1587,7 @@ Tags a given thread. The following restrictions apply:
 
 - Active and inactive chats can be tagged.
 - One or more tags can be added to the specific **thread**.
-- To tag a chat, the **tag** needs to be created first. It can be done in the [LiveChat Application](https://accounts.livechat.com/) or via the [Configuration API](/management/configuration-api).
+- To tag a chat, the **tag** needs to be created first. It can be done in the [LiveChat Application](https://accounts.livechat.com/) or via the [Configuration API](/management/configuration-api/v2.0#tags).
 - Tags are case sensitive.
 - Tags are defined by each client individually. The **list of tags** can be found in the [LiveChat Application](https://my.livechatinc.com/settings/tags). 
    
--- a/src/pages/messaging/agent-chat-api/index.mdx
+++ b/src/pages/messaging/agent-chat-api/index.mdx
@@ -1585,7 +1585,7 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
 
 Tags a given **thread**. 
 
-[Read how to list all available chats](https://my.livechatinc.com/settings/tags) and [create a new tag](/management/configuration-api/v2.0#add-a-tag). **Tags** are case sensitive in this method.
+Read how to [list all available chats](management/configuration-api/v2.0/#list-all-tags) and [create a new tag](/management/configuration-api/v2.0/#add-a-tag). **Tags** are case sensitive in this method.
 
 #### Specifics
 
--- a/src/pages/messaging/agent-chat-api/index.mdx
+++ b/src/pages/messaging/agent-chat-api/index.mdx
@@ -1585,7 +1585,7 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
 
 Tags a given **thread**. 
 
-Read how to [list all available tags](/management/configuration-api/v2.0#list-all-tags) and [create a new tag](/management/configuration-api/v2.0#add-a-tag). **Tags** are case sensitive in this method.
+Read how to [list all available tags](/management/configuration-api/v2.0#list-all-tags) and [create a new tag](/management/configuration-api/v2.0#add-a-tag).
 
 #### Specifics
 
@@ -1602,7 +1602,7 @@ Read how to [list all available tags](/management/configuration-api/v2.0#list-al
 | ----------- | -------- | --------- | ------------------------------------------ |
 | `chat_id`   | Yes      | `string`  | Id of the chat you want to add a tag to.   |
 | `thread_id` | Yes      | `string`  | Id of the thread you want to add a tag to. |
-| `tag`       | Yes      | `string`  | Tag name                                   |
+| `tag`       | Yes      | `string`  | Tag name. It's case sensitive.             |
 
 #### Response
 
@@ -1633,7 +1633,7 @@ https://api.livechatinc.com/v3.4/agent/action/tag_thread \
 
 ### Untag Thread
 
-Untags a given **thread**. **Tags** are case sensitive in this method.
+Untags a given **thread**.
 
 #### Specifics
 
@@ -1650,7 +1650,7 @@ Untags a given **thread**. **Tags** are case sensitive in this method.
 | ----------- | -------- | --------- | ----------------------------------------------- |
 | `chat_id`   | Yes      | `string`  | Id of the chat you want to remove a tag from.   |
 | `thread_id` | Yes      | `string`  | Id of the thread you want to remove a tag from. |
-| `tag`       | Yes      | `string`  | Tag name                                        |
+| `tag`       | Yes      | `string`  | Tag name. It's case sensitive.                  |
 
 #### Response
 
--- a/src/pages/messaging/agent-chat-api/index.mdx
+++ b/src/pages/messaging/agent-chat-api/index.mdx
@@ -1585,7 +1585,7 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
 
 Tags a given **thread**. 
 
-Read how to [list all available chats](management/configuration-api/v2.0/#list-all-tags) and [create a new tag](/management/configuration-api/v2.0/#add-a-tag). **Tags** are case sensitive in this method.
+Read how to [list all available chats](/management/configuration-api/v2.0#list-all-tags) and [create a new tag](/management/configuration-api/v2.0#add-a-tag). **Tags** are case sensitive in this method.
 
 #### Specifics
 
--- a/src/pages/messaging/agent-chat-api/index.mdx
+++ b/src/pages/messaging/agent-chat-api/index.mdx
@@ -1585,7 +1585,7 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
 
 Tags a given **thread**. 
 
-Read how to [list all available chats](/management/configuration-api/v2.0#list-all-tags) and [create a new tag](/management/configuration-api/v2.0#add-a-tag). **Tags** are case sensitive in this method.
+Read how to [list all available tags](/management/configuration-api/v2.0#list-all-tags) and [create a new tag](/management/configuration-api/v2.0#add-a-tag). **Tags** are case sensitive in this method.
 
 #### Specifics
 
--- a/src/pages/messaging/agent-chat-api/v3.5/index.mdx
+++ b/src/pages/messaging/agent-chat-api/v3.5/index.mdx
@@ -1580,7 +1580,7 @@ https://api.livechatinc.com/v3.5/agent/action/delete_event_properties \
 
 Tags a given **thread**. 
 
-Read how to [list all available chats](/management/configuration-api/v3.5#list-tags) and [create a new tag](/management/configuration-api/v3.5#create-tag). In this method, `tag` is case sensitive.
+Read how to [list all available tags](/management/configuration-api/v3.5#list-tags) and [create a new tag](/management/configuration-api/v3.5#create-tag).
 
 #### Specifics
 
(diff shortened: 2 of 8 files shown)


src/pages/messaging/agent-chat-api/index.mdx
```diff
@@ -1583,6 +1583,10 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
 
 ### Tag Thread
 
+To make a request, you need to specify the **tag name**, the **Id of the chat** and the **Id of the thread** you want to add a **tag** to. Active and inactive chats can be tagged, and one or more tags can be added to the specific **thread**. To tag a chat, the **tag** needs to be created first. It's possible to do that in the [LiveChat Application](https://accounts.livechat.com/) or via the [Configuration API](/management/configuration-api). Tags are case sensitive and are defined by each client individually. The **list of tags** can be found in the [LiveChat Application](https://my.livechatinc.com/settings/tags).
```
ale-kacperczyk, line 1586:
```suggestion
Active and inactive chats can be tagged, and one or more tags can be added to the specific **thread**. To tag a chat, the **tag** needs to be created first. It's possible to do that in the [LiveChat Application](https://accounts.livechat.com/) or via the [Configuration API](/management/configuration-api). Tags are case sensitive and are defined by each client individually. The **list of tags** can be found in the [LiveChat Application](https://my.livechatinc.com/settings/tags). 
```

ale-kacperczyk: We could shorten the description by deleting the first sentence. The same information is presented in the table below (request parameters). I suggest starting the description with a simple statement, for example:

> Tags a given thread.
>
> The rest of the description.

src/pages/messaging/agent-chat-api/index.mdx
```diff
@@ -1629,6 +1633,10 @@ https://api.livechatinc.com/v3.4/agent/action/tag_thread \
 
 ### Untag Thread
 
+To make a request, you need to specify the **tag name**, the **Id of the chat** and the **Id of the thread** you want to remove a **tag** from. Active and inactive chats can be untagged, and one or more tags can be removed from the specific **thread**. Tags are case sensitive.
+
+The **untagged threads** can be tagged using [Tag Thread](/messaging/agent-chat-api#tag-thread).
```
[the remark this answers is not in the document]
ale-kacperczyk: I would apply the same suggestions as for the Tag Thread method.

src/pages/messaging/agent-chat-api/index.mdx
```diff
@@ -1583,6 +1583,16 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
 
 ### Tag Thread
 
+Tags a given thread. The following restrictions apply:
+
+- Active and inactive chats can be tagged.
+- One or more tags can be added to the specific **thread**.
+- To tag a chat, the **tag** needs to be created first. It can be done in the [LiveChat Application](https://accounts.livechat.com/) or via the [Configuration API](/management/configuration-api).
+- Tags are case sensitive.
+- Tags are defined by each client individually. The **list of tags** can be found in the [LiveChat Application](https://my.livechatinc.com/settings/tags).
```
ale-kacperczyk, line 1592:
```suggestion
- Tags are defined by each customer individually. The **list of tags** can be found in the [LiveChat Application](https://my.livechatinc.com/settings/tags). 
```

ale-kacperczyk: Or maybe we should write "user"? WDYT, @oliwiapolec ?

src/pages/messaging/agent-chat-api/index.mdx
```diff
@@ -1583,6 +1583,16 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
 
 ### Tag Thread
 
+Tags a given thread. The following restrictions apply:
+
+- Active and inactive chats can be tagged.
+- One or more tags can be added to the specific **thread**.
+- To tag a chat, the **tag** needs to be created first. It can be done in the [LiveChat Application](https://accounts.livechat.com/) or via the [Configuration API](/management/configuration-api).
```
ale-kacperczyk, line 1590:
```suggestion
- To tag a chat, the **tag** needs to be created first. It can be done in the [LiveChat Application](https://accounts.livechat.com/) or via the [Configuration API](/management/configuration-api/v2.0#tags).
```

src/pages/messaging/agent-chat-api/index.mdx
oliwiapolec: From the mentioned I think we should lean towards "user" (or maybe license, or admin? as only admins can alter the tags), we very rarely use "client" in the copy and the "customer" would be rather connoting the end-user

kamilatokarska: So I think, in that case, I would go with 'admit' if only they could alter the tags. WDYT @ale-kacperczyk?

ale-kacperczyk: My only objection to "admin" is that nowhere else are we so specific with roles in method descriptions. But 'user' might confuse readers to think that tags cannot be globally set by a license, but need to be created by each user. So, let's go with 'admin' and see what a reviewer from the API team says :)

src/pages/messaging/agent-chat-api/index.mdx
```diff
@@ -1583,6 +1583,16 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
 
 ### Tag Thread
 
+Tags a given thread. The following restrictions apply:
```
oliwiapolec, line 1586: I'm wondering whether we should call the below list "restrictions" as they're more "possibilities"? 🤔 E.g. for the threads, there's rather no restrictions as a thread can be pretty much only either active or inactive :D, and another: tags are case sensitive is not rather a restriction but a fact, you know what I mean :D

src/pages/messaging/agent-chat-api/index.mdx
oliwiapolec, line 1590:
```suggestion
- To tag a chat, the **tag** needs to be created first. It can be done in the [LiveChat Application](https://my.livechatinc.com/) or via the [Configuration API](/management/configuration-api).
```

oliwiapolec: Or have I missed something and we redirect to the Accounts panel instead of the AA @ale-kacperczyk? 😅

src/pages/messaging/agent-chat-api/index.mdx
```diff
@@ -1629,6 +1639,14 @@ https://api.livechatinc.com/v3.4/agent/action/tag_thread \
 
 ### Untag Thread
 
+Untags a given thread. The following restrictions apply:
+
+- Active and inactive chats can be untagged. 
+- One or more tags can be removed from the specific **thread**. 
+- Tags are case sensitive.
```
oliwiapolec, line 1646: Hmm, I'm thinking if this belongs here, as we know already that tags are case sensitive from the previous "Tag thread" description and it's more of a tag-specific thing rather than method-specific information 🤔

src/pages/messaging/agent-chat-api/index.mdx
kamilatokarska: The word 'restrictions' I based on the fact that that is how the intro sentences to the different bullet lists were introduced. They did not have the same info. However, some were similar https://developers.livechat.com/docs/messaging/agent-chat-api/rtm-reference#chat-users
However, I get your point, so maybe - ...Attention should be paid to the following information."?

src/pages/messaging/agent-chat-api/index.mdx
kamilatokarska: I put that rather 'just to be sure', but now that I am thinking about what you have written, I am more for removing it from here.

ale-kacperczyk: Removing it from this description would assume someone has written the Tag Thread method description. I they haven't and we remove the description, one might not one that calling Untag Thread with `support` and `Support` is removing two separate tags.

src/pages/messaging/agent-chat-api/index.mdx
ale-kacperczyk: What about "rules"?

src/pages/messaging/agent-chat-api/index.mdx
kamilatokarska: Good point, so this should definitely stay.

src/pages/messaging/agent-chat-api/index.mdx
kamilatokarska: With 'rules', I think I have similar worries like Oliwia above with 'restrictions' - '...Active and inactive chats can be tagged..' or '...The tags’ list can be found in the LiveChat  Application.' - it more sounds like a fact/info. 
So maybe 
"...Attention should be paid to the following rules and information:..."?
or
"The following rules and information apply..." (If we want to follow the rest of the bullet lists' intros)

I cannot come up with one word describing all the points...

gluszczyk: "terms and conditions apply" 😂 :trollface: 

Can we drop this "preface" altogether and just do a new "paragraph" with all this additional info?

src/pages/messaging/agent-chat-api/index.mdx
```diff
@@ -1583,6 +1583,16 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
 
 ### Tag Thread
 
+Tags a given thread. The following restrictions apply:
+
+- Active and inactive chats can be tagged.
```
gluszczyk, line 1588: let's drop this
you tag threads, not tags (just as the method name suggests 😉 )
and threads can be either active or not, there is no in-between, it's just like saying "you can tag all threads"

src/pages/messaging/agent-chat-api/index.mdx
```diff
@@ -1583,6 +1583,16 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
...
+- One or more tags can be added to the specific **thread**.
```
gluszczyk, line 1589: this method allows adding a single tag in a single call so maybe it's better to emphasize this as a thread property? something like "a thread can have multiple tags"?

src/pages/messaging/agent-chat-api/index.mdx
gluszczyk, line 1590: livechatINC ? someone needs "the talk" :trollface: 
anyway, I would avoid pushing the user towards an app when they are browsing API docs 😉

gluszczyk, line 1590: livechatINC ? someone needs "the talk" :trollface: 
anyway, I would avoid pushing the user towards an app when they are browsing API docs 😉

gluszczyk: livechatINC ? someone needs "the talk" :trollface: 
anyway, I would avoid pushing the user towards AA when they are browsing API docs 😉

src/pages/messaging/agent-chat-api/index.mdx
gluszczyk: I would combine point starting with "To tag a chat..." with this one and go with
"you can list available tags here [link]... to create a new tag check/see [link]"
so there would be no mention of the user (and in case a "user" has to be mentioned then it should be admin, as only admins have sso-scopes that allow creating tags)

also pls drop
>To tag a chat, the **tag** needs to be created first.

this is like saying you need to create a thread first, otherwise, what will you tag? kinda obvious 😉

src/pages/messaging/agent-chat-api/index.mdx
```diff
@@ -1583,6 +1583,16 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
...
+- To tag a chat, the **tag** needs to be created first. It can be done in the [LiveChat Application](https://accounts.livechat.com/) or via the [Configuration API](/management/configuration-api).
+- Tags are case sensitive.
```
gluszczyk, line 1591: It's ok here but I would consider skipping it in 3.5

src/pages/messaging/agent-chat-api/index.mdx
```diff
@@ -1583,6 +1583,16 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
...
+- Tags are defined by each client individually. The **list of tags** can be found in the [LiveChat Application](https://my.livechatinc.com/settings/tags). 
+   
+The **tagged threads** can be untagged using [Untag Thread](#untag-thread).
```
gluszczyk, line 1594: I don't think we mention opposite methods in other places (eg remove_user in add_user description) 🤔 

anyway, if we are going to do so, let's go with "to remove an existing tag from the thread use..."
"tagged thread can be untagged" sounds... meh

src/pages/messaging/agent-chat-api/index.mdx
gluszczyk: one more thing, we should update "chat" data structure and add "tags" array to the thread there

src/pages/messaging/agent-chat-api/index.mdx
kamilatokarska: It is not crucial, so I will remove it :)

src/pages/messaging/agent-chat-api/index.mdx
ale-kacperczyk: It is obvious when you know it, but there are people who will try to use this method and forget that they need to create a tag first. How about:

> To tag a chat, the tag needs to be [created](/management/configuration-api/v3.5#create-tag) first.

This way, we will kill two birds with one stone (reminding people that they need to create a tag first & redirect them to the new tag methods).

src/pages/messaging/agent-chat-api/index.mdx
ale-kacperczyk: chats -> threads 👍🏻
As for active and inactive, I specifically remember @bkubiak mentioning it as a valuable addition to this description

src/pages/messaging/agent-chat-api/index.mdx
ale-kacperczyk: Well, actually, https://my.livechatinc.com/ still has "inc". Joke's on you XD

src/pages/messaging/agent-chat-api/index.mdx
kamilatokarska: I would stick to the bullet list as it will be easier for the reader to go through all this info. Also, each point is different, so I believe a bullet list would be more effective.

src/pages/messaging/agent-chat-api/index.mdx
kamilatokarska: For me, it is not as obvious as it may seem to be, I would stick to this info...I believe that with inactive threads it may not be so obvious for the user that he/she can also tag them - I would suggest changing to: "Active and inactive threads can be tagged" or "You can tag active and inactive threads"

src/pages/messaging/agent-chat-api/index.mdx
kamilatokarska: If I understand correctly - What do you think to write it directly - One tag can be added in one call, but one thread can have multiple tags'? 🤔 

"chat" data structure and add "tags" array to the thread there - I will create a task for that in Jira and investigate 😄

src/pages/messaging/agent-chat-api/index.mdx
kamilatokarska: Just to make sure if I get it right - you mean skipping both points completely?
To confirm - tags will not be case sensitive in 3.5, yes?

And also...I cannot find this comment but you wrote that you would avoid pushing the user toward AA when they are browsing API docs - Can I ask why cause I was not aware of that 😅

src/pages/messaging/agent-chat-api/index.mdx
gluszczyk: I didn't even consider that I need to verify it, and yet you are right, AA still uses "inc", good grief, "kajam się" XD

gluszczyk: but still, I don't think this is the place to mention AA; you check how to tag a thread with API, and to add a new tag we send you to this GUI app... meh

src/pages/messaging/agent-chat-api/index.mdx
gluszczyk: still, some points are related and can be grouped as separate "paragraphs"

src/pages/messaging/agent-chat-api/index.mdx
gluszczyk: idk about @bkubiak's opinion but maybe you should add him to this review instead :trollface: 
for me this states "any thread" in a different way, and most of these methods are usable on any thread unless specified otherwise

src/pages/messaging/agent-chat-api/index.mdx
gluszczyk: no need for a separate task 😉 just add it [here](https://github.com/livechat/livechat-public-docs/blob/master/payloads/messaging/v3.4/agent-chat-api/other-structures/chat.json) cc @ale-kacperczyk 

>One tag can be added in one call, but one thread can have multiple tags

idk, the interface kinda forces you to provide a single tag, and I don't like stating obvious things :P

src/pages/messaging/agent-chat-api/index.mdx
gluszczyk: this comment was only about case sensitivity... and in 3.5 it's more complicated, there is another PR that describes changes there so maybe the description in 3.5 should be based on that

as for AA aversion... if a developer comes to create their own integration and reads API docs we should show him how to do things API-way, not asking him to use "some GUI app" to skip API requests

src/pages/messaging/agent-chat-api/index.mdx
gluszczyk: that's why I wanted to combine listing and creating tags in a single point... as for "tag needs to be created first", it's not an API-obvious but a life-obvious kinda thing

ale-kacperczyk: You would think that it's life-obvious that you first need to register a property and only then are you able to update it, yet some users still try to update a non-existent property and look for property creation options in the Agent Chat API. The same might be true for tags and linking the method that allows for creating would help tackle that.

Okay, but I think merging the info about available tags & how to create tags into a single point is a nice compromise. 

In v3.2-v3.4, I would add:

Read how to [list all available tags](/management/configuration-api/v2.0#list-all-tags) and [create a new tag](/management/configuration-api/v2.0/#add-a-tag).

In v3.5, I would add a similar note but link to v3.5 methods instead.

kamilatokarska: I would go with the compromise as I am also not so sure if it would be life-obvious for everyone.

gluszczyk: >Read how to [list all available tags](https://github.com/management/configuration-api/v2.0#list-all-tags) and [create a new tag](https://github.com/management/configuration-api/v2.0/#add-a-tag).

@ale-kacperczyk this is exactly what I had in mind 😉 
and my remark about obviousness is not about putting the above info in but about the "the tag needs to be created first" sentence... like duh

ale-kacperczyk: OK, I get what you mean 👍🏻

src/pages/messaging/agent-chat-api/index.mdx
ale-kacperczyk: Plus we can replace "restrictions" with "info" if needed.

kamilatokarska: Yes, I was thinking to change 'restrictions' for 'infrmation'

src/pages/messaging/agent-chat-api/index.mdx
kamilatokarska: I don't know.... 😅  I still support my opinion that for some people possibility of adding tags to inactive threads may not be so obvious. 
@ale-kacperczyk WDYT? Maybe I just measure people's intelligence too low

src/pages/messaging/agent-chat-api/index.mdx
ale-kacperczyk: Since we agreed to merge two points into one, this entire point can be deleted.

src/pages/messaging/agent-chat-api/index.mdx
ale-kacperczyk: 1. It's a dilemma between "Write it explicitly in the description so that people know right away" and "Give people a minute to analyze request parameters so that they figure out what can be sent themselves".

> One tag can be added in one call, but one thread can have multiple tags

"One tag can be added in one call" can be concluded from the request parameters, so we would be restating what's already expressed in the table. That leaves us with the information that "a thread can have multiple tags", which is true, but would be more fitting for a general overview of tags (like the PR @gluszczyk prepared for v3.5). It's only loosely related to this method.

Taking all into account, I think that maybe it's best to remove this line altogether?

2. We will update the chat data structure in a separate PR right after we're done with this one. That's just a more convenient way to do it for us.

src/pages/messaging/agent-chat-api/index.mdx
ale-kacperczyk: Concerning case sensitivity:

1. Keep the "- Tags are case sensitive." sentence in v3.2-v3.4.
2. As for v3.5, my assumption is that the v3.5 Tag Thread method checks if a tag already exists before tagging a thread, so we should adjust the v3.5 description. Something like, "In this method, letter case needs to be respected for tag names." This would suggest that it's only for this method, and not all tag-related methods in v3.5. 
@gluszczyk, correct me if my story's made up :)

src/pages/messaging/agent-chat-api/index.mdx
ale-kacperczyk: I wanted to add him for counter-balance but he's not available :) 

Jokes aside, I would continue to argue that this would be a helpful addition to novice developers, but this 👇 forces me to capitulate :)

> most of these methods are usable on any thread unless specified otherwise

@kamilatokarska, let's drop this line. Despite our good intentions, keeping the info could create some potential confusion as to why for some methods we specify that it works for active and inactive threads, and for other methods, we don't.

src/pages/messaging/agent-chat-api/index.mdx
ale-kacperczyk: It looks like the description will shrink after all the corrections :) Let's see if it still makes sense to keep the bullet list. If there are at least 3 bullet points, then I think it's worth it. If fewer, then it's better to format it as a paragraph.

src/pages/messaging/agent-chat-api/index.mdx
kamilatokarska: Yes, I agree that if someone analyzes the request parameters then the fact that 'one tag can be added in one call' can be figured out. 
I would say that the info that 'a thread can have multiple tags' is important, but you may be right that it fits more the general overview of tag.
I will remove that info completely:)

src/pages/messaging/agent-chat-api/index.mdx
kamilatokarska: Both methods will have less than 3 bullet points, so it makes no sense to keep the bullet list.  Paragraphs will be better in that case.

src/pages/messaging/agent-chat-api/index.mdx
kamilatokarska: Already deleted:)

src/pages/messaging/agent-chat-api/index.mdx
kamilatokarska: I have pushed new changes with the v3.5 as I think it will be better to discuss when seeing it:)

src/pages/messaging/agent-chat-api/index.mdx
```diff
@@ -1583,6 +1583,10 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
 
 ### Tag Thread
 
+Tags a given **thread**. 
+
+[Read how to list all available chats](https://my.livechatinc.com/settings/tags) and [create a new tag](/management/configuration-api/v2.0#add-a-tag). **Tags** are case sensitive in this method.
```
kamilatokarska, line 1588: 'create a new tag' link is not linking to the appropriate part...I have checked it and it is correct, so I do not why it is happening :/

src/pages/messaging/agent-chat-api/v3.5/index.mdx
```diff
@@ -1578,6 +1578,10 @@ https://api.livechatinc.com/v3.5/agent/action/delete_event_properties \
 
 ### Tag Thread
 
+Tags a given **thread**. 
+
+[Read how to list all available chats](https://my.livechatinc.com/settings/tags) and [create a new tag](/management/configuration-api/v2.0#add-a-tag). In this method, letter case needs to be respected for tag names.
```
kamilatokarska, line 1583: 1. For now, 'create a new tag' is linked to the v3.2 as I believe the v3.5 has not been published yet. Please correct me if I am wrong.
2. Please let me know if this information about the letter case is enough.

src/pages/messaging/agent-chat-api/v3.5/index.mdx
```diff
@@ -1624,6 +1628,8 @@ https://api.livechatinc.com/v3.5/agent/action/tag_thread \
 
 ### Untag Thread
 
+Untags a given **thread**. In this method, letter case needs to be respected for tag names.
```
kamilatokarska, line 1631: I am not sure if I understood correctly xd in this method letter case needs to be respected as well?

src/pages/messaging/agent-chat-api/index.mdx
gluszczyk: @ale-kacperczyk in v3.5, **Create Tag** is the first method that ignores letter case when checking if the tag already exists, and the sentence "Tags are case sensitive" becomes **mostly** true 😉 so it might be better to state case-sensitivity explicitly in tag methods

maybe the sentence can explicitly state which field needs to respect the letter case?
>In this method, `tag` is case sensitive.

Grammarly fixed case sensitivity into case-sensitivity 😉 so what's with this BrE vs AmE you talked about in the other PR? 🤔

src/pages/messaging/agent-chat-api/v3.5/index.mdx
gluszczyk: yes, both tag and untag methods are case-sensitive

src/pages/messaging/agent-chat-api/index.mdx
ale-kacperczyk: @gluszczyk It may be due to your Grammarly settings (maybe, idk). I had consulted "case sensitive vs case-sensitive" with our in-house editor and he recommended to stick to "case sensitive" without a hyphen.

src/pages/messaging/agent-chat-api/v3.5/index.mdx
ale-kacperczyk: 1. Could you please link to v3.5? We will first release the v3.5 PR, and then yours, so that the link works. I suggested a link change.
2. We could go with the sentence proposed by @gluszczyk "In this method, `tag` is case sensitive." It better because it mentions the `tag` field explicitly.

src/pages/messaging/agent-chat-api/index.mdx
ale-kacperczyk, line 1588:
```suggestion
Read how to [list all available chats](management/configuration-api/v2.0/#list-all-tags) and [create a new tag](/management/configuration-api/v2.0/#add-a-tag). **Tags** are case sensitive in this method.
```

src/pages/messaging/agent-chat-api/index.mdx
```diff
@@ -1583,6 +1583,10 @@ https://api.livechatinc.com/v3.4/agent/action/delete_event_properties \
 
 ### Tag Thread
 
+Tags a given **thread**. 
+
+Read how to [list all available chats](/management/configuration-api/v2.0#list-all-tags) and [create a new tag](/management/configuration-api/v2.0#add-a-tag). **Tags** are case sensitive in this method.
```
ale-kacperczyk, line 1588:
```suggestion
Read how to [list all available tags](/management/configuration-api/v2.0#list-all-tags) and [create a new tag](/management/configuration-api/v2.0#add-a-tag). **Tags** are case sensitive in this method.
```

ale-kacperczyk, line 1588: How about we move the info about case sensitivity to the Request table to comply with the format proposed in [this PR](https://github.com/livechat/livechat-public-docs/pull/1145/files#diff-26505ab33f8366eff9fa237c91538cc37a2f61c11cdd44e89ba26aeb8779d7b5R2492)?
image

ale-kacperczyk: The same should be applied in other places

kamilatokarska: Yes, I like the idea! Just to confirm - all the places, including v3.5?

gluszczyk (approved):

ale-kacperczyk (approved):
👏🏻👏🏻👏🏻

PR #1167, opened 2022-06-09
Remove firebase notifications.

gluszczyk:
# Description

removed content applies to old mobile apps which are no longer supported


gluszczyk:
branch naming fail 😂

PR #1170, opened 2022-06-10
Remove firebase notifications.

gluszczyk:
# Description

removed content applies to old mobile apps which are no longer supported


bkubiak:
Hey, shouldn't we inform devs about the removal of the method in the changelog? So that the devs can adjust their code during the migration process to the higher version?

ale-kacperczyk:
> Hey, shouldn't we inform devs about the removal of the method in the changelog? So that the devs can adjust their code during the migration process to the higher version?

I don't know. We discussed it with @gluszczyk and came to the conclusion that there's no need to mention it in the CHangelog since it hasn't been functional in any version we currently have documented publicly.

PR #1220, opened 2022-08-29
DED-276 linking sample apps in Sign in with LiveChat

kamilatokarska:
# Deploy preview

# Links

Link your Jira ticket.

- [Jira]()

# Description

# Review and release

### DIFF
--- a/src/pages/authorization/sign-in-with-livechat/index.mdx
+++ b/src/pages/authorization/sign-in-with-livechat/index.mdx
@@ -320,6 +320,10 @@ A popup object instance could be also created via `sdk` with:
 
 - [authorize](#authorize)
 
+**Sample app with the popup flow**
+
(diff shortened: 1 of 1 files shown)


src/pages/authorization/sign-in-with-livechat/index.mdx
```diff
@@ -320,6 +320,10 @@ A popup object instance could be also created via `sdk` with:
 
 - [authorize](#authorize)
 
+**Sample app with the popup flow**
```
ale-kacperczyk, line 323: Not sure if this is the right place for the link – we're separating "Available class methods" and the method description/sample. I would add the link to the sample app at the end of this section. The same for Redirect

kamilatokarska: I was actually split between those two places for it. I will try to put it at the end of this section as you are right that separating "Available class methods" and the method description/sample may be disturbing.

PR #1340, opened 2023-03-28
New SEO meta tags

ale-kacperczyk:
# Deploy preview

# Description
- Added support for the `robots` meta tag. We need to mark several documents as `noindex`.
- Added support for a subtitle that will only be displayed in meta tags, not on the page. This subtitle is optional.

### DIFF
--- a/src/components/core/seo.js
+++ b/src/components/core/seo.js
@@ -11,7 +11,7 @@ const defaultSiteMetadata = {
   author: "@livechat",
 };
 
-function SEO({ desc, keywords, title, subtitle }) {
+function SEO({ desc, keywords, title, subtitle, robots }) {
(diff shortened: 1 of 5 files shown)


src/components/core/seo.js
```diff
@@ -11,16 +11,17 @@ const defaultSiteMetadata = {
   author: "@livechat",
 };
 
-function SEO({ desc, keywords, title }) {
+function SEO({ desc, keywords, title, subtitle, robots }) {
```
arturfracala, line 14: You didn't add these props to PropTypes. Check the validity of passed props to avoid problems in the future.

arturfracala (changes requested):

PR #1359, opened 2023-04-19
API-12676: fix scopes description in Add User To Chat

litwinow:
# Links

- [Jira](https://livechatinc.atlassian.net/browse/API-12676)

# Description

Fix scopes description in Add User To Chat.
Scope `chats--all:rw` allows access to all chats only for the requester, so when adding other agents using token with this scope, they still have to have access to that chat.
Scope `chats--access:rw` requires the user's access, not the requester's access to match the chat access.


kamilatokarska (approved):
Thanks for the info:)

PR #1404, opened 2023-06-22
API-12939: Clarify when to use enable/disable license webhooks

litwinow:
# Links

- [Jira](https://livechatinc.atlassian.net/browse/API-12939)

# Description

Clarify when to use Enable/Disable License Webhooks.

### DIFF
--- a/src/pages/management/configuration-api/v3.6/index.mdx
+++ b/src/pages/management/configuration-api/v3.6/index.mdx
@@ -2804,7 +2804,7 @@ Here's what you need to know about **webhooks**:
 
 ### Register Webhook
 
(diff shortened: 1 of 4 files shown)


bkubiak (approved):
All clear now, thanks

src/pages/management/configuration-api/v3.6/index.mdx
```diff
@@ -3059,9 +3059,10 @@ curl -X POST \
 
 ### Enable License Webhooks
```
kamilatokarska, line 3060: I was wondering - in all the other versions, there is also a change in the Register Webhook method - shouldn't it be made in this version as well?

litwinow: had the changes locally, forgot to save the file 🤦

kamilatokarska (approved):

PR #1465, opened 2024-02-29
Moved Chat API to Text API category

ale-kacperczyk:


src/pages/text/team-chat-api/index.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+## FAQ
+
+### Is Team Chat API the same as Agent Chat API?
+
+No. Team Chat API is a separate service independent of the LiveChat product.
+
+### Can I use Team Chat API to chat with customers?
+
+Team Chat API was designed for building custom chat applications for internal communication (agent - agent; team member - team member).
```
bkubiak, line 37: > (agent - agent; team member - team member)
I don't understand why we use both names together. Perhaps we should also mention bots here because it's one of the main use-cases, wdyt?

src/pages/text/team-chat-api/index.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+### Can I build apps for LiveChat with Team Chat API?
+
+It's not designed for building apps for LiveChat. If you're looking for a solution to chat with customers, you should use the [Agent Chat API](https://developers.livechat.com/docs/messaging/agent-chat-api/) and [Customer Chat API](https://developers.livechat.com/docs/messaging/customer-chat-api/).
```
bkubiak, line 41: Actually, this is not 100% true. This API is designed to be independent from LiveChat, but an app for LiveChat can definitely use this API.

bkubiak (commented):
I like the new place more (and especially the name that reflects its purpose).

I think we should also tweak the name messaging APIs, currently being used for agent-api and customer-api to reflect that it's meant for business chat with customers.

src/pages/text/team-chat-api/index.mdx
ale-kacperczyk: I used both names to give more examples because I feel like "agent" is very business-oriented while "team member" better reflects the internal communication use case. Sure, I can add bots

src/pages/text/team-chat-api/index.mdx
ale-kacperczyk: You're right. I meant something else, I need to rephrase it.

src/pages/text/team-chat-api/index.mdx
bkubiak: Yes but I mean that using "team member" after ";"  looks like it is something different. How about putting "team member" in brackets?

bkubiak (approved):

PR #1469, opened 2024-03-11
Add docs for payment_trialstarted app webhooks

Kaweckip:
# Links

Link your Jira ticket.

- [Jira](https://livechatinc.atlassian.net/browse/MP-1658)

# Description
Adds new payment trial started webhook.


olek-p (commented):
I can see the actual webhook's payload also contains organization ID and user ID

Kaweckip:
Yeap, I've noticed that as well. Probably we should update all the webhooks then

ale-kacperczyk (approved):

PR #1575, opened 2025-07-15
feat: agent-api: change field name from last_visit to visit in customer object v3.6

wfabjanczuk:
Jira: https://livechatinc.atlassian.net/browse/CDP-348


oliwiapolec (approved):
DM me if we can merge 😌

